PR #30, opened 2022-02-04
feat: Mumble 3d positional audio

maximegmd:
Needs to be tested, I am not sure of my math


Merlin6019:
Did some testing, and this is what I found:

- Skyrim Together successfully links with mumble (link in plugin list shows Skyrim Together).
image

- When successfully linked and positional audio is set up properly on both ends, and both players are connected to the server, neither can hear the other.
image

- If one player disconnects from the game server, that player will go back to global voice and audio in the channel and can speak to and hear each player connected to the server separately, but players that are connected to the server can not hear other players connected to the game server (but can speak to and hear the player disconnected from the game server).

- 1 unit = 1.428 cm, not 1 meter (https://www.creationkit.com/index.php?title=Unit).

- Even if the maximum distance is set to 500 in positional audio instead of 50, still no audio. I suspect something isn't adding up with player positions.

PR #60, opened 2022-03-09
New inventory system

RobbeBryssinck:
This new system builds up a custom inventory with all relevant extra data (as far as I was able to enumerate it). This means individual items can now also be manipulated by the server. The inventories and their items can be exposed to a server side scripting API.

The system could be further optimized by adding and removing individual items on change instead of sending over the entire inventory every time it changes, but for testing purposes, I will do that for the next test build. Custom names for items are not synced yet, since this was causing some crashes, but I can add this at a later date.

This new system should fix 90% of the crashes that the playtesters are experiencing.


Luc45:
Dumb question: Is it hard to add the capability of sharing crafted potions?

PR #202, opened 2022-06-19
Waypoints and player markers

JonahWeinbaum:
The player stores a queue of dummy map markers off-map on connection. When a new player joins, the map marker is moved from off-screen to the player's position. 

Custom destinations are handled similarly, although upon closing the map a real waypoint gets placed in the same location so that it appears on all player's compasses.


RobbeBryssinck:
(Partially) implemented in #613 .

PR #226, opened 2022-07-08
disable geolocation in the serverlist for now

Dragonisser:
> The geolocation code relies on a thirdparty service that rate limits requests, this is a temporary work around as it fails listing servers because there are too many. ~yamashi

### DIFF
--- a/Code/skyrim_ui/package-lock.json
+++ b/Code/skyrim_ui/package-lock.json
@@ -1,13 +1,17427 @@
 {
   "name": "skyrim-together-ui",
   "version": "0.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "skyrim-together-ui",
+      "version": "0.0.0",
+      "dependencies": {
+        "@angular/animations": "^11.2.3",
+        "@angular/cdk": "^11.2.13",
+        "@angular/common": "^11.2.3",
+        "@angular/compiler": "^11.2.3",
+        "@angular/core": "^11.2.3",
+        "@angular/forms": "^11.2.3",
+        "@angular/material": "^11.2.13",
+        "@angular/platform-browser": "^11.2.3",
+        "@angular/platform-browser-dynamic": "^11.2.3",
+        "@angular/router": "^11.2.3",
+        "@fortawesome/angular-fontawesome": "^0.8.2",
+        "@fortawesome/fontawesome-svg-core": "^1.2.15",
+        "@fortawesome/free-regular-svg-icons": "^5.9.0",
+        "@fortawesome/free-solid-svg-icons": "^5.7.2",
+        "rxjs": "~6.5.5",
+        "tslib": "^2.0.0",
+        "zone.js": "~0.10.3"
+      },
+      "devDependencies": {
+        "@angular-devkit/build-angular": "^0.1102.2",
+        "@angular/cli": "^11.2.2",
+        "@angular/compiler-cli": "^11.2.3",
+        "@types/jasmine": "^3.6.4",
+        "@types/jasminewd2": "^2.0.8",
+        "@types/node": "^12.11.1",
+        "codelyzer": "^6.0.0",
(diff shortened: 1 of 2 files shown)


Sebdevar:
Any explanations concerning the removal?

yamashi:
The geolocation code relies on a thirdparty service that rate limits requests, this is a temporary work around as it fails listing servers because there are too many.

Sebdevar (approved):
I'd perhaps add a mention of the issue and necessary fix as part of the commented code ;)

danidomen:
As the PR is good to disable the geolocation in order to have the server list working, the best solution is that the part to get the geodata for every server is better to be done when a server start and then sent directly to the server list storage (https://skyrim-reborn-list.skyrim-together.com/list), but not that every client make calls to ip-api.com (https://github.com/tiltedphoques/TiltedEvolution/blob/1e9bbf09948d844386e13adcc4003208ef8eadae/Code/skyrim_ui/src/app/components/server-list/server-list.component.ts#L96
) because that client will get the rate limit in a few queries and stop the execution, as for now are 1443 servers on the list :D

Dragonisser:
> As the PR is good to disable the geolocation in order to have the server list working, the best solution is that the part to get the geodata for every server is better to be done when a server start and then sent directly to the server list storage (https://skyrim-reborn-list.skyrim-together.com/list), but not that every client make calls to ip-api.com (
> 
> https://github.com/tiltedphoques/TiltedEvolution/blob/1e9bbf09948d844386e13adcc4003208ef8eadae/Code/skyrim_ui/src/app/components/server-list/server-list.component.ts#L96
> 
> 
> ) because that client will get the rate limit in a few queries and stop the execution, as for now are 1443 servers on the list :D
> Also, thinking more about this, is better to be server variables to be defined in the .ini, and only be at default sended into the list pinning with the ip-api.com if not setted. Why? Because for example, we have the server in a german cloud provider, but on the list we don't want to appear as "Germany", as we are a Spanish fellows, so that is better to be defined by the server manager.

The idea was to show the location of the server to know where it is (latency), not the location of the person that hosts it, since we still have no way in showing the latency of said servers.

Having the server send its location is still a better idea than doing it on the clientside.

Dragonisser:
> > > As the PR is good to disable the geolocation in order to have the server list working, the best solution is that the part to get the geodata for every server is better to be done when a server start and then sent directly to the server list storage (https://skyrim-reborn-list.skyrim-together.com/list), but not that every client make calls to ip-api.com (
> > > https://github.com/tiltedphoques/TiltedEvolution/blob/1e9bbf09948d844386e13adcc4003208ef8eadae/Code/skyrim_ui/src/app/components/server-list/server-list.component.ts#L96
> > > 
> > > ) because that client will get the rate limit in a few queries and stop the execution, as for now are 1443 servers on the list :D
> > > Also, thinking more about this, is better to be server variables to be defined in the .ini, and only be at default sended into the list pinning with the ip-api.com if not setted. Why? Because for example, we have the server in a german cloud provider, but on the list we don't want to appear as "Germany", as we are a Spanish fellows, so that is better to be defined by the server manager.
> > 
> > 
> > The idea was to show the location of the server to know where it is (latency), not the location of the person that hosts it. Reason why is simply since we still have no way in showing the latency of said servers.
> > Having the server send its location is still a better idea than doing it on the clientside.
> 
> That is a good idea, about to know the latency between the server and the user, but then something like https://www.npmjs.com/package/net-ping in the client to make a short request to the server to know the latency in ms between them could be a valid approach?. I mean, if the server is located in Spain and I'm in Spain could have better latency, but not always is true because of the infra of the server or the internet capability of the client. Anyway, moving the definition of that variables of "country", "countryCode", "continent" to the server startup is the best way as your proposal in #247 :)
> 
> Or maybe, as the server sents the info to the /list api, something not a lot of workload on the clientside could be send the ping between the server and the list server, so at least could be a more reasonable timming to know what servers are "fast" (I know, that will be raising a latency between the server location against the list api server, but less is nothing, and maybe with that you know what servers are the most "fasters")


There is no node backend.

PR #243, opened 2022-07-10
Add reference to bug report repository in Readme

croscas:
Note that personally I am not sure that it is worth to keep a separate repository just for bugs, however, as long as it is the case it is important to make it visible.


RobbeBryssinck:
Make pull requests to dev, not master.

croscas:
@RobbeBryssinck done.

RobbeBryssinck:
Awesome, this is good info, let's hope people will read it :D

PR #253, opened 2022-07-11
feat: Toggle drop sync server setting

Merlin6019:


croscas:
Can you update the description of the PR with meaningful information? For example what problem is it solving, the changes required in the codebase and how have you tested that this change works as expected.

RobbeBryssinck:
The idea is good, but there are some implementation problems. This is rather overkill. We're sending the option to the client here, but that seems unnecessary. The client doesn't really have to do anything for this behavior to function. Furthermore, we don't really need to store it in ServerSettings in this case. It can just stand on its own. Look at fGoldLossFactor for a better example.

Code/server/Services/InventoryService.cpp
```diff
@@ -10,6 +10,13 @@
 #include <Messages/RequestEquipmentChanges.h>
 #include <Messages/NotifyEquipmentChanges.h>
 #include <Messages/DrawWeaponRequest.h>
+#include <client/World.cpp>
```
RobbeBryssinck, line 13: Same thing, don't include cpp.

Code/server/Services/InventoryService.cpp
```diff
@@ -36,7 +41,11 @@ void InventoryService::OnInventoryChanges(const PacketEvent<RequestInventoryChan
     NotifyInventoryChanges notify;
     notify.ServerId = message.ServerId;
     notify.Item = message.Item;
-    notify.Drop = message.Drop;
+
+    if (bEnableItemDrops)
```
yamashi, line 45: These 4 lines could just be

notify.Drop = bEnableItemDrops ? message.Drop : false;

yamashi (changes requested):

PR #261, opened 2022-07-12
Automatic player home detection for no-sync purposes

ejektaflex:
This PR allows Skyrim Together Reborn to check whether a given cell is a player home automatically so that it can disable sync of chest inventories. This allows STR to work with almost all mods that add custom player homes, and it also reduces the reliance on hardcoded IDs.

We detect player homes by looking for cells that contain the `NoResetZone` encounter zone. Despite the name (and the fact that many zones may have a flag that disallows cell resets), the *only* cells that have this encounter zone are player homes, with the exception of Sinderion's Field Laboratory, which has been coded in as an exception (though, one could argue that Sinderion is never going to visit, and it's about as close to a home that you ever might find in Blackreach).


Code/client/Games/Skyrim/Forms/LoadedCellData.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include <Components/BaseFormComponent.h>
+
+struct LoadedCellData
+{
+    uint8_t pad160[0x160]; // Offset
```
RobbeBryssinck, line 7: For padding naming, we usually do "pad" and then the current offset, so that'd be "pad0" here.

Code/client/Games/Skyrim/Forms/LoadedCellData.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
```
RobbeBryssinck, line 1: LoadedCellData is not a form. It is part of TESObjectCELL, so you should move the struct to be within TESObjectCELL.

Code/client/Services/Generic/ObjectService.cpp
```diff
@@ -44,19 +46,25 @@ ObjectService::ObjectService(World& aWorld, entt::dispatcher& aDispatcher, Trans
 #endif
 }
 
-bool IsVanillaHome(const uint32_t acCellId) noexcept
+bool IsPlayerHome(const TESObjectCELL* pCell) noexcept
 {
-    switch (acCellId)
+
+    if (pCell->loadedCellData->encounterZone)
```
RobbeBryssinck, line 52: Check if loadedCellData is null first

RobbeBryssinck (changes requested):

Code/client/Services/Generic/ObjectService.cpp
yamashi: As well as pCell, can never be too careful

ejektaflex:
I've made these changes, renaming the padding, moving the struct out of Forms/ and into the cell form, and doing extra null checks!

PR #265, opened 2022-07-12
Make XP sync toggleable through the server

RobbeBryssinck:


Code/server/Services/CharacterService.cpp
```diff
@@ -43,6 +43,11 @@
 #include <Messages/NotifyActorTeleport.h>
 #include <Messages/NotifyRelinquishControl.h>
 
+namespace
+{
+Console::Setting bEnableXpSync{"Gameplay:fGoldLossFactor", "Syncs combat XP within the party", true};
```
Malivil, line 48:
```suggestion
Console::Setting bEnableXpSync{"Gameplay:bEnableXpSync", "Syncs combat XP within the party", true};
```

Malivil (changes requested):

Code/server/Services/CharacterService.cpp
RobbeBryssinck: woops, im a noob, thanks

Malivil: Happens to the best of us =)

Malivil (approved):

PR #293, opened 2022-07-13
Cyrillic UI font

Sidiusz:
Adds Cyrillic support to the font
Before the font had "standard windows characters" for Russian letters, I added letters in the style of the original font
Also now you can write in chat in Russian and it does not turn into a "mess" of symbols


croscas:
Can you update the description explaining why this change is needed and what does it fix?

Sidiusz:
Added

> Can you update the description explaining why this change is needed and what does it fix?

Dragonisser:
Please get rid of every other change than the font. Make a new branch based on the up-to-date **dev** and PR this then.

Dragonisser (approved):
Can be merged

PR #318, opened 2022-07-15
Some more UI changes

scomans:
- add notifiactions on party invites (click to open party menu)
- increased size of serverlist buttons
- fix row height in server list on ultra wide screens
- add little animation to the health bar when the value changes
- add max player count to server list
- adjusted server list
- added custom drowdowns
- added language selection
- make checkboxes look like from Skyrim

image

PR #322, opened 2022-07-15
Fix #321 UI scale for ultrawide resolution (7680x1440p)

### LINKED ISSUE #321: UI Doesn't Scale Correctly for Ultrawide Resolution (7680x1440p)
## Description
Server browser is squished and "Join" buttons don't work when playing on 7680x1440p resolution. In the 1.0 release, this was not the case. I would consider this a low priority as I understand this resolution is not in widespread use. Please fix other more pressing issues first.

## Steps to reproduce
1. Start the game.
2. Load a save.
3. Open the server browser.

## Reproduction rate
[ X] Consistent: I can reproduce this bug by following the steps to reproduce 100% of the time (*every time*)

## Expected result
For the UI to not be squished and join buttons to work.

## Evidence (optional)
server-menu

Dumbeldor:


Dumbeldor:
I only have 1080p screens but I was able to reproduce the problem by faking the resolution in firefox.
 It should be fine

scomans:
You can't do that. It will break the virtual scrolling of the list.

Dumbeldor:
Aye @scomans are you sure about that ?
image
It seems ok

scomans:
It will look fine in the middle but every row with more than 38 pixel will result in entries being cut off at the end of the list.

Dragonisser:
Sorry Dumbo :P 
https://github.com/tiltedphoques/TiltedEvolution/pull/318

Issue #110, opened 2022-04-16
Client chat commands


RobbeBryssinck:
- [x] Teleport to another player
- [ ] List of players in server with their names (thought maybe make this a UI component?)
- [ ] Whisper

PR #324, opened 2022-07-15
feat: Teleport chat command

Sturlen:
Lets you use `/teleport PLAYER_NAME` in chat to teleport to a player. Player name is case-insensitive and only compares base letters, e.g. typing in "jorgen" will match a player named "Jørgen". Also supports spaces.

Planning to implement rest the chat features requested in Issue #110 at a later date.


Dragonisser:
Player name can have spaces

Dragonisser:
Also, its already implemented 

https://github.com/tiltedphoques/TiltedEvolution/blob/master/Code/client/Services/Generic/CommandService.cpp

Sturlen:
Did not realize the C++ side already handled that, good to know!

Dragonisser:
> Did not realize the C++ side already handled that, good to know!

To be fair, i forgot about it as well.

PR #330, opened 2022-07-16
Add config option for player home / NoCellReset sync

### LINKED ISSUE #308: Request for Chest contents synching in dlc/modded homes to be a toggleable feature on servers
## Description

A toggleable server function for whether or not storage in dlc or modded housing syncs between players.

## Additional context

The 1.1 patch turns this off completely, but this can cause issues with housing that's set to dynamically update depending on what's been placed in storage. This is fundamentally how the storage in Legacy of the Dragonborn functions. If this isn't allowed to sync between players then the features of mods such as this will not sync between players either.

ejektaflex:
For most players, leaving player home sync disabled is fine. However, some players might prefer to have their home base synced for various reasons. 

This adds a new config variable, `bSyncPlayerHomes`, defaulting to `false`, which toggles whether player homes are synced.

Legacy of the Dragonborn is one such example - it has it's own player home, but contains functionality that depends on the contents of several chests within the home. As such, a player might want to re-enable sync if they want to play with it.

This fixes #308!

### DIFF
--- a/Code/client/Services/Generic/ObjectService.cpp
+++ b/Code/client/Services/Generic/ObjectService.cpp
@@ -96,7 +96,7 @@ void ObjectService::OnCellChange(const CellChangeEvent& acEvent) noexcept
 
     // Player homes should not be synced, so that chest contents,
     // which are often used as storage, are never accidentally wiped.
-    if (IsPlayerHome(pCell) && !World::Get().GetServerSettings().SyncPlayerHomes)
+    if (!World::Get().GetServerSettings().SyncPlayerHomes && IsPlayerHome(pCell))
         return;
 
     GameId cellId{};
--- a/Code/client/Services/Generic/ObjectService.cpp
+++ b/Code/client/Services/Generic/ObjectService.cpp
@@ -96,7 +96,7 @@ void ObjectService::OnCellChange(const CellChangeEvent& acEvent) noexcept
 
     // Player homes should not be synced, so that chest contents,
     // which are often used as storage, are never accidentally wiped.
-    if (IsPlayerHome(pCell))
+    if (IsPlayerHome(pCell) && !World::Get().GetServerSettings().SyncPlayerHomes)
         return;
 
     GameId cellId{};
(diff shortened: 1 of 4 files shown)


Dragonisser:
Syncing by default is probably the better idea since vanilla is what we intend the mod for.

ejektaflex:
That is true, though it might inadvertently mess with many people's existing servers when they upgrade. That being said, if we have to set a precedent, we may as well do it now. I've made a commit changing the default to `true`.

Dragonisser:
It was already set to true anyway since 1.1.0 came with it forced to true. As of now it wont change much when the update drops with default enabled. It just needs to be documented and explained in the announcement.

RobbeBryssinck:
Syncing by default means that the contents in player home chests are unsafe, and might get overwritten. This is a bad idea.

Code/server/GameServer.cpp
```diff
@@ -36,6 +37,7 @@ Console::StringSetting sPassword{"GameServer:sPassword", "Server password", ""};
 Console::Setting uDifficulty{"Gameplay:uDifficulty", "In game difficulty (0 to 5)", 4u};
 Console::Setting bEnableGreetings{"Gameplay:bEnableGreetings", "Enables NPC greetings (disabled by default since they can be spammy with dialogue sync)", false};
 Console::Setting bEnablePvp{"Gameplay:bEnablePvp", "Enables pvp", false};
+Console::Setting bSyncPlayerHomes{"Gameplay:bSyncPlayerHomes", "Sync player homes and other NoResetZones", true};
```
RobbeBryssinck, line 40: Should be false by default.

Code/client/Services/Generic/ObjectService.cpp
```diff
@@ -96,7 +96,7 @@ void ObjectService::OnCellChange(const CellChangeEvent& acEvent) noexcept
 
     // Player homes should not be synced, so that chest contents,
     // which are often used as storage, are never accidentally wiped.
-    if (IsPlayerHome(pCell))
+    if (IsPlayerHome(pCell) && !World::Get().GetServerSettings().SyncPlayerHomes)
```
RobbeBryssinck, line 99: It's more efficient to check for the SyncPlayerHomes variable before calling IsPlayerHome

RobbeBryssinck (changes requested):

ejektaflex:
I've reverted it and made those changes. Thanks!

Code/server/GameServer.cpp
```diff
@@ -36,6 +37,7 @@ Console::StringSetting sPassword{"GameServer:sPassword", "Server password", ""};
 Console::Setting uDifficulty{"Gameplay:uDifficulty", "In game difficulty (0 to 5)", 4u};
 Console::Setting bEnableGreetings{"Gameplay:bEnableGreetings", "Enables NPC greetings (disabled by default since they can be spammy with dialogue sync)", false};
 Console::Setting bEnablePvp{"Gameplay:bEnablePvp", "Enables pvp", false};
+Console::Setting bSyncPlayerHomes{"Gameplay:bSyncPlayerHomes", "Sync player homes and other NoResetZones", false};
```
RobbeBryssinck, line 40: I'd change the description to "Sync chests and displays in player homes and other NoResetZones"

Issue #329, opened 2022-07-15
temporary HP boost extending outside of the UI frame


French-Asgeir:
## Description

Didn't get to test for other form of HP boost (potion or enchant) but arkay blessing which give +25hp seems to extend outside of the default HP frame for group member.

Related formIDs|
---------------|

none

## Steps to reproduce

1. Start the game.
2. Load a saved game.
3. Connect to the server.
4. Create a party
5. Ask a friend to connect and join your party
6. You and you friend use any shrine of Arkay.
7. You should now see the HP bar on the top left corner extend out of it's designed frame

## Reproduction rate

Please check just _one_ of the following reproduction rates:

- [x] Consistent: I can reproduce this bug by following the steps to reproduce 100% of the time (*every time*)
- [ ] Occasional: I can reproduce this bug by following the steps to reproduce around 50% of the time
- [ ] Rare: I can reproduce this bug by following the steps to reproduce around 10% or less of the time

## Expected result

4 Options : 

![image](https://user-images.githubusercontent.com/109287366/179319469-34ce6d80-32cc-4ea8-a8b5-f1ef65e4972c.png)

Option 2 or 4 would be the best one since it probably won't cause weird behavior when using an absurdly good HP potion (which is easy to do in skyrim

## Evidence (optional)

Forgot the screenshot while playing sorry. The paint above transcript the issue really well

Dragonisser:
Even tho the calculation is basically just health / maxhealth * 100, it seems as maxhealth isn't including this temporary buff.

Dragonisser:
Someone would need to RE to find the actual maxHealth value

![image](https://user-images.githubusercontent.com/4233773/179355184-8f1729b2-f081-41ff-bd03-8208d9b4359d.png)
![image](https://user-images.githubusercontent.com/4233773/179355201-42cffcb7-fb9b-4593-90ff-383a8e21c62e.png)


```
float CalculateHealthPercentage(Actor* apActor) noexcept
{
    ActorValueOwner& actorValueOwner = apActor->actorValueOwner;

    spdlog::info("Sending health information");
    spdlog::info("ActorValueOwner");
   

    const float actorHealth = actorValueOwner.GetValue(24);
    const float actorBaseHealth = actorValueOwner.GetBaseValue(24);
    const float actorPermaHealth = actorValueOwner.GetPermanentValue(24);

    spdlog::info("actorHealth: {}", actorHealth);
[...]

PR #336, opened 2022-07-16
Fix: Health Bar Overflow

kulcris:
Fix for health bars overflowing past the UI.
This simply clamps the value to 100% so that it does not overflow. Numbers over 100% are ignored and treated as 100%
If max health is 100
Player has 150 health this will read 100% until player is under 100 (max health)
Player has 35 health this will read 35%

2nd time's the charm...?

### DIFF
--- a/Code/client/Services/Generic/OverlayService.cpp
+++ b/Code/client/Services/Generic/OverlayService.cpp
@@ -102,6 +102,8 @@ float CalculateHealthPercentage(Actor* apActor) noexcept
     float percentage = health / maxHealth * 100.f;
     if (percentage < 0.f)
         percentage = 0.f;
+	if (percentage > 100.f)
+		percentage = 100.f;
 
     return percentage;
 }


kulcris:
Fixes this issue
https://github.com/tiltedphoques/TiltedEvolution/issues/329

RobbeBryssinck:
This is more of a hacky workaround, not a fix. The root problem here is that the percentage should never be above 100 anyway. The calculation is flawed, and this should be addressed instead.

Vlado-VG:
That doesn't really make much sense. You can still have the health bar displaying a % while the health being between 100 and 150. Assume 150hp is 100%. 100/150 would be 66.6% health.
You can have the health displayed as x/150, while the red bar displaying the percentage.

RobbeBryssinck:
> That doesn't really make much sense. You can still have the health bar displaying a % while the health being between 100 and 150. Assume 150hp is 100%. 100/150 health points would be 66.6%. You can have the health displayed as x/150 in numbers, while the red progress bar displaying the percentage.

Not sure who you're responding to here or what you're trying to say.

Dragonisser:
As cosi said, the calculations are bugged and fixing it like this isnt a clean solution.

Vlado-VG:
> Not sure who you're responding to here or what you're trying to say.

I was responding to kulcris. You are correct that calculation is flawed.

Dragonisser:
The calculations need to be addressed, so make a fix for that and create a new PR

Issue #326, opened 2022-07-15
Restore enemy health on player kill


MatthewTeply:
## Description

For additional challenge, a new option would be available when creating a server, "Restore enemy health on player kill". This would make it so when a player dies (host, or any other player), the enemy that killed that player would gain it's health back. This would prevent you from killing mammoths at level 2, because you keep spawning around them and stabbing them with iron daggers.

sicsix:
Could include an option respawning the player with low health / magicka as well

DJLegends1011:
nice implementation

French-Asgeir:
being able to chose the amount of health restored (a percentage of max health) could be a nice addition

DLumi:
Maybe not if one player got killed, but the whole party got wiped out? 
I mean, if just one of the party members got killed and the enemy got health replenished - that's gonna be pretty sad for the dudes who survived.

MatthewTeply:
@French-Asgeir - This is going to be a feature, percentage of HP restored is 100% (pun intended) going to be an option.

@DLumi - I'm currently working on implementation, and it is not really feasible to track if a single enemy wiped out the entire party, I'm sure it is possible somehow, but it would be a lot more complex. I'm looking into an option to restore the enemy's health based on party size.

ejektaflex:
I floated the idea earlier of restoring their health by (1/party size)% when a player is killed. E.g. when there are five players, and one dies, all nearby enemies regain 20% (1/5) of their health.

The main concern with this idea was that it may be unfair if some characters are not contributing (e.g. if it's two players and one is only a healer), but honestly I don't think it's too big a deal.

xForcer:
This feature was never implemented in the end? 
I'll go ahead and try making something simple, with configurable options!

Rolothe5th:
there id mods that does this similar

PR #337, opened 2022-07-16
Reduced player Health and Magicka on Respawn

sicsix:
Drops the player down to 10% health and magicka after exiting god mode during respawn.
Could help reduce respawn zerg rushing somewhat and get players to retreat and regroup instead.

Is this worth making into a server variable?

### DIFF
--- a/Code/client/Services/Generic/PlayerService.cpp
+++ b/Code/client/Services/Generic/PlayerService.cpp
@@ -235,6 +235,15 @@ void PlayerService::RunPostDeathUpdates(const double acDeltaTime) noexcept
         {
             PlayerCharacter::SetGodMode(false);
 
+            PlayerCharacter* pPlayer = PlayerCharacter::Get();
+            
+            // A little punishment for dying to prevent respawn rushing
+            // Should this multiplier be a server var?
+            float spawnHealth = pPlayer->GetActorPermanentValue(ActorValueInfo::kHealth) * 0.1f;
+            float spawnMagicka = pPlayer->GetActorPermanentValue(ActorValueInfo::kMagicka) * 0.1f;
+            pPlayer->ForceActorValue(ActorValueOwner::ForceMode::DAMAGE, ActorValueInfo::kHealth, spawnHealth);
+            pPlayer->ForceActorValue(ActorValueOwner::ForceMode::DAMAGE, ActorValueInfo::kMagicka, spawnMagicka);
+
             godmodeStart = false;
         }
     }
--- a/Code/client/Services/Generic/PlayerService.cpp
+++ b/Code/client/Services/Generic/PlayerService.cpp
@@ -238,11 +238,11 @@ void PlayerService::RunPostDeathUpdates(const double acDeltaTime) noexcept
             PlayerCharacter* pPlayer = PlayerCharacter::Get();
             
             // A little punishment for dying to prevent respawn rushing
-            // Should this multiplier be a server var?
-            float spawnHealth = pPlayer->GetActorPermanentValue(ActorValueInfo::kHealth) * 0.1f;
-            float spawnMagicka = pPlayer->GetActorPermanentValue(ActorValueInfo::kMagicka) * 0.1f;
-            pPlayer->ForceActorValue(ActorValueOwner::ForceMode::DAMAGE, ActorValueInfo::kHealth, spawnHealth);
-            pPlayer->ForceActorValue(ActorValueOwner::ForceMode::DAMAGE, ActorValueInfo::kMagicka, spawnMagicka);
+            float respawnFactor = World::Get().GetServerSettings().HealthAndMagickaOnRespawnFactor;
+            float health = pPlayer->GetActorPermanentValue(ActorValueInfo::kHealth) * respawnFactor;
+            float magicka = pPlayer->GetActorPermanentValue(ActorValueInfo::kMagicka) * respawnFactor;
+            pPlayer->ForceActorValue(ActorValueOwner::ForceMode::DAMAGE, ActorValueInfo::kHealth, health);
+            pPlayer->ForceActorValue(ActorValueOwner::ForceMode::DAMAGE, ActorValueInfo::kMagicka, magicka);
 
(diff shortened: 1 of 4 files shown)


imniko:
Would def love this as a server variable, like setting it between 0.1-1.0

RobbeBryssinck:
I don't know whether this is a good solution. When I played with my friends, and we fought a dragon in the overworld, I just kept respawning nearby until we killed the dragon. Having low health on respawn would simply mean I'd respawn more, but the result would be the same.

The gold loss factor is already a pretty good solution for those who find death to not be punishing enough.

sicsix:
https://github.com/tiltedphoques/TiltedEvolution/issues/326 also has another idea, resetting the health of an NPC that kills a player, which might work well with this.

I think there needs to be more disincentives to respawn rushing.

imniko:
Losing a fixed % of gold (maybe configurable?) ala borderlands game could also be a good option.

RobbeBryssinck:
That already is an option, look at fGoldLossFactor in the server config.

sicsix:
Gold loss could make it expensive provided you are carrying gold, reduced health and magicka makes it impractical and perhaps NPC health resetting makes it impossible?

In single player when you die you might reload and try the encounter again. Currently you 'reload' from halfway through the encounter and it feels a bit like cheating.

Sturlen:
Adding a stacking damage reduction after death would make infinite revive less viable. Would need a way to reset this after the encounter though.

Dragonisser:
> Gold loss could make it expensive provided you are carrying gold, reduced health and magicka makes it impractical and perhaps NPC health resetting makes it impossible?
> 
> In single player when you die you might reload and try the encounter again. Currently you 'reload' from halfway through the encounter and it feels a bit like cheating.

Reloading a save isn't an option tho because its desyncs a lot. So this respawning wont change, only what happens after respawning can be adjusted imo. If its damage reduction, gold loss, enemy get their health back, etc.

sicsix:
I might look into if and how to handle restoring NPC health on a player kill separately to this since it opens a can of worms. Determining the damage source may be tricky so I don't think it'd be possible or practical to restore health to the killer but there are other methods that might work, such as restoring health to all loaded enemy NPCs of each party member? Would need to have a good think about it.

Bar any issues or suggestions I think this PR is done.

MatthewTeply:
> I might look into if and how to handle restoring NPC health on a player kill separately to this since it opens a can of worms.
> 
> Bar any issues or suggestions I think this PR is done.

I'm currently working on this in a seperate issue: #326

RobbeBryssinck:
My main problem with this PR still remains, in that it does not help game balance or difficulty. This just makes it slightly more tedious to enter the fight again, since you just have to wait to respawn and walk 40 meters again to continue fighting. Eventually, you could still kill that dragon. Now, if the dragon itself regenerated (part of) its own health on player death, for example, then there's a real possibility that you may never be able to kill the dragon, and thus have to flee, which makes death have meaning.

sicsix:
The argument would best be made from the point of view where health and magicka have their regeneration rates changed be it from the vanilla atronach stone or difficulty mods. These require spending resources to regenerate, such as potions etc. Dying right now regenerates those resources instantly and for free, taking all the difficulty out of managing them.
It also helps deal with zerging. Yes you could argue that it's more tedious to enter the fight again but isn't that the point? If dying is too easy with no downsides there's no fear of dying.

RobbeBryssinck:
The point of death shouldn't be to make it more tedious, it should be to make it more difficult. Ultimately, the result is still the same: I can still kill the dragon if I just keep respawning. There is no difference to me respawning with reduced health. The outcome will always be the same, it's just a bit tedious to have to walk back a bit. Contrast that to the dragon receiving (parts of) its health back, and suddenly, there's real consequence, since we might never be able to kill the dragon if we keep dying. In other words, there's a possibility for different outcomes now. That's what adds difficulty beyond just being mildly tedious.

sicsix:
I'm coming from the perspective of mods like Requiem where regenerating all your health and magicka for free is a major advantage and ruins any feeling of challenge in group play. I understand it may not make as much difference in vanilla since the regeneration rate is much higher.

The other perspective too is not just fighting a dragon, but say a large group of bandits. We've had it before where the respawn occurs within a very short distance of the bandits and you come back into the fight with full health for free. A lone bandit regenerating his health won't make much difference to the outcome here.

If enabling this option for everyone isn't desired, maybe there could be an event fired that scripts could subscribe to to allow them to make adjustments on respawn?

RobbeBryssinck:
> The other perspective too is not just fighting a dragon, but say a large group of bandits. We've had it before where the respawn occurs within a very short distance of the bandits and you come back into the fight with full health for free. A lone bandit regenerating his health won't make much difference to the outcome here, and it really just turns into a bunch of players winning fights above their pay grade by zerging the enemy to death.

With your proposed death system, they can do exactly that though. They just keep respawning nearby, and they can keep zerging them, as they do right now. The only difference is that they, annoyingly, have to get up and run back to the fight more often. The outcome is still the same: the enemy bandits get zerged. It doesn't add difficulty, just slight annoyance. Contrast that with another death system where the enemy regains their health instead, and there suddenly is a real possibility that you and your friends might not be able to defeat the enemy, hence creating different potential outcomes, which creates true difficulty instead of just a mild annoyance to achieve the same outcome.

> If enabling this option for everyone isn't desired, maybe there could be an event fired that scripts could subscribe to to allow them to make adjustments on respawn?

We would need a client side scripting API first. Papyrus is not enough, since we cannot send out custom events.

sicsix:
I think the very act of making something more difficult and time consuming to do is a valid method of stopping that behaviour.
If I had to respawn dozens of times to win a battle due to coming back in with low health versus only a handful of time with full health, well personally that's enough to make me not want to continue, but everyone is different!

JacobLouis:
I am almost wondering if instead of reducing health and magika it would be better to have a respawn timer as punishment. So if you die you would have to wait a configurable amount of time before respawning. If you really wanted to go nuts you could have talking to a downed player trigger a instant respawn.

PR #339, opened 2022-07-16
Fallout Together WIP port

RobbeBryssinck:
This PR merges the ported Reborn code into Fallout Together, and makes Fallout Together build and run again. This is NOT a release for Fallout Together. This just makes it so that people have a chance to work on it again, if they wish to do so. When making contributions to the project in the future, the PR must compile for both Skyrim and Fallout Together.

Fallout Together in its current state compiles, runs, and is able to take multiple players. It has some bugs, and needs Fallout 4 specific features to be synchronized. Check the "issues" board for issues with the "Fallout" label for more elaboration on those issues.


RobbeBryssinck:
Skyrim Together Reborn should be tested with this branch. Since this is such a massive PR, it is possible that something in Reborn might have broken during the port, since the services were affected.

Sturlen:
I've built and run both ST and FT from this branch and haven't found any issues so far.

RobbeBryssinck:
> I've built and run both ST and FT from this branch and haven't found any issues so far.

To what extent did you test ST? And mostly, this will require some static analysis.

PR #340, opened 2022-07-16
Fix: UI not showing up

RobbeBryssinck:
This should be a more definitive fix.


RobbeBryssinck:
Seems to work most of the time, but still rarely happens.

Issue #368, opened 2022-07-20
Can not send player more than one party invite per party.


dancomstock:
## Description

When inviting a player to a party, if they don't or can't accept the invite, they can not be sent a new one without first disbanding the whole party and starting a new one.

In my opinion, this is partially a bug and partially a feature request as I believe some UI changes are also needed.  I will create a separate issue for the ui changes I believe are needed to fully resolve the underlying issue, and then link here.

## Steps to reproduce

1. Invite a player to a party.
2. Invited player either is already in a party or accidentally clicks on the option to create their own party.
3. No option to invite player again, invite button does nothing until party is disbanded and a new party is created.

## Reproduction rate

- [x] Consistent: I can reproduce this bug by following the steps to reproduce 100% of the time (*every time*)
- [ ] Occasional: I can reproduce this bug by following the steps to reproduce around 50% of the time
- [ ] Rare: I can reproduce this bug by following the steps to reproduce around 10% or less of the time

## Expected result

Able to reinvite a player to a party if they don't accept the first time.  Not having to disband the party and reinvite everyone.

SaintMercury:
An easy solution at the moment could be to not disallow "spam inviting" of another player; basically, not disable the invite button after sending an invite.

Another solution could be to debounce the invite on a per player basis, but unsure if that's a desired functionality, since by the sounds of it, the project's goal is to be low player count servers.

PR #371, opened 2022-07-21
DRAFT: Issue 368: Can not send player more than one party invite per party. 

SaintMercury:
https://github.com/tiltedphoques/TiltedEvolution/issues/368

remove 'hasBeenInvited' check in list

### DIFF
--- a/Code/client/Services/Generic/OverlayService.cpp
+++ b/Code/client/Services/Generic/OverlayService.cpp
@@ -15,6 +15,7 @@
 #include <Services/TransportService.h>
 
 #include <Messages/NotifyChatMessageBroadcast.h>
+#include <Messages/NotifyPlayerPartyLeft.h>
 #include <Messages/NotifyPlayerList.h>
 #include <Messages/NotifyPlayerLeft.h>
 #include <Messages/NotifyPlayerJoined.h>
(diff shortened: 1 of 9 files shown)


Dragonisser:
1. Wrong branch, its dev
2. That's not a fix for the issue. Find the actual problem, which would be resetting the value for the player when they joined any party.

SaintMercury:
@Dragonisser 

Updated to be a less hacky solution, though I'm not married to this one. This properly resets the flag set on a client when a player leaves a party.

Dragonisser:
That wont work. `PartyService::OnPartyLeft` is for the local player. So what you need to do is send that information to the server, which then sends it to everyone else. **Then** you can send that information to the UI to reset the flag.

Code/skyrim_ui/src/app/services/player-list.service.ts
```diff
@@ -120,6 +123,18 @@ export class PlayerListService {
     });
   }
 
+  private onPlayerPartyLeave() {
+    this.playerPartyLeaveSubscription = this.clientService.partyLeftChange.subscribe((leavingPlayerId: number) => {
+      const playerList = this.getPlayerList();
+
+      if (playerList) {
+        this.getPlayerById(leavingPlayerId).hasInvitedLocalPlayer = false;
```
SaintMercury, line 131: need to null check here, since a player could've just left the game, not necessarily a party

Code/server/Services/PartyService.h
```diff
@@ -50,6 +50,7 @@ struct PartyService
 
     void BroadcastPlayerList(Player* apPlayer = nullptr) const noexcept;
     void BroadcastPartyInfo(uint32_t aPartyId) const noexcept;
+    void BroadcastPlayerLeave(Player* apPlayer = nullptr) const noexcept;
```
SaintMercury, line 53: default `nullptr` is probably unnecessary

SaintMercury:
I've since updated this branch to use a new server message type.

My understanding is as follows:
When a player leaves a party, from the server's perspective, notify other players of the leaving players id (and party leaving).
Clients listen for this new specific message, and reset the flag accordingly.

Dragonisser:
I guess this has been fully tested in game?

SaintMercury:
Not yet, I've been busy lately. I can get it tested by monday probably.

SaintMercury:
This does **NOT** work atm. I get a crash when a player leaves the party. Need to keep working on it :/

Dragonisser:
Stale PR

Issue #331, opened 2022-07-16
Font Size Settings


Sturlen:
## Description

Default text size can be difficult to read on small screens. Would be nice to be able to set font size in the settings menu, even if it's just a couple presets.

## Additional context

Should only need to change the font-size at the root element and it should propagate down, based on what I've seen of the codebase.

Dragonisser:
For that we need to get the dropdown and scrollbar to work with the mouse.

Sturlen:
Could use buttons to increase and decrease size then. Skyrim already has a similar menu control:
<img width="155" alt="difficulty" src="https://user-images.githubusercontent.com/43607012/179469095-deb36af9-c09d-48e2-abfa-9c5582d64d51.png">
<img width="155" alt="font size input" src="https://user-images.githubusercontent.com/43607012/179469112-d69e829b-ef7a-49ac-ae82-794cfcfff815.png">

Dragonisser:
Could be an option, though the issue mentioned needs to be fixed anyway.

Sturlen:
Very true, but I think it might fit better in a separate bug report.

Dragonisser:
Just saying :P

PR #373, opened 2022-07-21
Change font size via Settings menu

Sturlen:
Fix for #331 
Font size is adjustable between 10 and 20 px and is saved as any other setting. 
Had to make changes to all SCSS files which used a static calculation of rem, otherwise they won't scale correctly. Could probably go above 20px, but it would require more UI changes.

bilde
"Font" and "Font size" are both translateable.


Code/skyrim_ui/src/app/components/root/root.component.ts
```diff
@@ -81,6 +84,12 @@ export class RootComponent implements OnInit {
       });
   }
 
+  public onFontSizeSubscription() {
+    this.settings.fontSizeChange.subscribe( size => {
```
scomans, line 88: Unsubscription missing (takeUntil)

Code/skyrim_ui/src/app/components/settings/settings.component.html
```diff
@@ -112,6 +112,19 @@ <h2>{{ 'COMPONENT.SETTINGS.DEBUG' | transloco }}</h2>
         >
       </div>
     </div>
+    <div class="section">
+      <h2>{{ 'COMPONENT.SETTINGS.FONT' | transloco }}</h2>
+      <div class="flex-item">
+        <label for="font-size">{{ 'COMPONENT.SETTINGS.FONT_SIZE' | transloco }}</label>
+        <span class="fontsize">
+          <button id="decrement" (click)="onFontSizeDown()"></button>
+            <input id="font-size" type="number" min=10 value={{fontSize}} max=20 step=1 [ngModel]="fontSize"
```
scomans, line 121: Use binding instead of interpolation

Code/skyrim_ui/src/app/components/settings/settings.component.scss
```diff
@@ -9,7 +9,6 @@
 
 .container {
   overflow: auto;
-  max-height: 60vh;
```
scomans, line 12: Why did you make this change?

Code/skyrim_ui/src/app/components/settings/settings.component.html
scomans: Why an input?

Code/skyrim_ui/src/app/components/settings/settings.component.ts
```diff
@@ -116,6 +121,23 @@ export class SettingsComponent implements OnInit {
     this.settingsUpdated.next();
   }
 
+  onFontSizeChange(size: number) {
+    if (size >= this.minFontSize && size <= this.maxFontSize) {
+      console.log("set")
```
scomans, line 126: Remove

Code/skyrim_ui/src/app/components/settings/settings.component.scss
Sturlen: When the font size increased you would need to use the the scrollbar, which currently do not work. This makes sure you can actually reach the font size option.

Code/skyrim_ui/src/app/components/settings/settings.component.ts
Sturlen: console.log removed

Dragonisser:
I'm honestly not really fond of those changes. This might need some discussing and utilising of more a dynamic approach.

https://stackoverflow.com/a/54708418

Code/skyrim_ui/src/app/components/settings/settings.component.html
Sturlen: Replaced with a text element now since it's just supposed to dispaly a value. Used input before because I'm not very familiar with Angular.

Sturlen:
Not very familiar Anuglar so I used the normal JS approach. I'll see what I can do with ngstyle.

Sturlen:
I've done some digging and I haven't found any other good ways to set the font-size on the root elelemt.
Ngstyle and ngclass wont work since we need to modify an element which is in`index.html`, which is outside the scope of Angular.

I also tried making a fontsize component with the selector  `html`, this worked it setting the font-size but it stopped rendering the rest of the app.

Setting the root text size is normally done with browser settings, not through code. That is not really an option here. 
This means the solution will be hacky, since most normal websites and framework have no good reason to do this.

Dragonisser:
> I've done some digging and I haven't found any other good ways to set the font-size on the root elelemt. Ngstyle and ngclass wont work since we need to modify an element which is in`index.html`, which is outside the scope of Angular.
> 
> I also tried making a fontsize component with the selector `html`, this worked it setting the font-size but it stopped rendering the rest of the app.
> 
> Setting the root text size is normally done with browser settings, not through code. That is not really an option here. This means the solution will be hacky, since most normal websites and framework have no good reason to do this.

You might wanna look how angular works. There is only app-root you would have to work with. Even then, creating themes and changing the classlist on a root level would also work.

image

Sturlen:
Rem units are based on the font size of the root html element. Setting font size on the app-root will have no effect on this.
Creating a theme would work if you were only setting a static value such as the background color, but here we need to dynamically set a value based on user input.

In theory you could create a class for every possibe size, but this would be extremely labour intensive and hard to maintain.
I'm aware that editing the root element is something that is NEVER done in a normal website, but this is an UI overlay embedded in an application, not a normal website.

Dragonisser:
> Rem units are based on the font size of the root html element. Setting font size on the app-root will have no effect on this. Creating a theme would work if you were only setting a static value such as the background color, but here we need to dynamically set a value based on user input.
> 
> In theory you could create a class for every possibe size, but this would be extremely labour intensive and hard to maintain. I'm aware that editing the root element is something that is NEVER done in a normal website, but this is an UI overlay embedded in an application, not a normal website.

I dont see why we even need to be able to set the font-size from 10px - 20px. It should be enough to have 3-4 options to choose from.

Sturlen:
It's absolutely possible, but it would require writing more code just to give players fewer options.

Stanoja:
Hi @Sturlen @Dragonisser, I did the font-size preset adjustments https://github.com/Sturlen/TiltedEvolution/pull/1

Main change that I did was instead of setting `document.style` we're working with the `rootComponent` as host.
Basically, we set `font-size-X` to the the host and the rest of the fixes are to make sure size affects children components.

I'm not proficient in css, so if you have a better approach let me know. :)

Sturlen:
@Stanoja I see where you're coming from, but I don't think adding 68 lines of code to this PR just to work around a standard browser feature is the way to go. Pretty much every large website, including google, facebook, and youtube, uses rem units that based on the document font size for a good reason.

If you really want to go for presets instead, just use `document.style ` to set your desired font size, based on which option the user selected from the settings panel.

Stanoja:
Good point @Sturlen.

I reverted everything related to the styling and also the font-size application, so that leaves this PR only with the Styling option.

Cheers!

Sturlen:
@Stanoja After testing your changes , I agree that having a a set of option is easier to understand than just picking a number.

Changes have been tested in game and are working as intended.

Code/skyrim_ui/src/app/components/settings/settings.component.scss
scomans: But without the max-height the overflow will not trigger and when the container gets too big (or the screen too small) it just gets cut at the bottom

Sturlen: Yes, I see your point now. I'll add it back with enough room for the font options.

scomans: Scrolling with the scrollwheel works fine so it should be reachable anyways. But putting it above the Debug section would be a good idea

Code/skyrim_ui/src/app/components/settings/settings.component.html
```diff
@@ -112,6 +112,17 @@ <h2>{{ 'COMPONENT.SETTINGS.DEBUG' | transloco }}</h2>
...
       </div>
     </div>
+    <div class="section">
+      <h2>{{ 'COMPONENT.SETTINGS.FONT' | transloco }}</h2>
+      <div class="flex-item">
+        <label for="fontSize">{{ 'COMPONENT.SETTINGS.FONT_SIZE' | transloco }}</label>
+        <span class="fontsize-controls">
+          <button id="decrement" (click)="onFontSizeDown()"></button>
+            <span id="fontSize">{{('COMPONENT.SETTINGS.FONT_SIZES.' + fontSize.toUpperCase()) | transloco}}</span>
+          <button id="increment"(click)="onFontSizeUp()"></button>
```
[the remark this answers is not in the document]
Sturlen: increment id is used for styling, the other one can be removed.

Code/skyrim_ui/src/app/components/settings/settings.component.scss
Sturlen: I think moving it ot the top would be best, since it's probably the one of the first things you want to change.

Sturlen:
Settings menu can no longer go outside the screen and fontsize is easily reachable.
Tested for 720p, 1080p and 4k.

Sturlen:
Now compatible with the massive new UI changes.

Sturlen:
Font size now fits in better with the rest of the settings UI and uses the new dropdown.
bilde

Sturlen:
@Dragonisser Unless there's any issues I'm not aware of, this should be ready to merge.

RobbeBryssinck:
@scomans could you perhaps review the UI part of this PR again? Dragonisser isn't available for reviews at the moment.

scomans:
Code looks good and it works

PR #375, opened 2022-07-22
Let UI handle connect/disconnect messages

Sturlen:
Currently the Client sends a system message with  "successfully connnected/disconnected" whenever you connect to or disconnect from a server. These strings are hardcoded in the Client and cannot be translated. Displaying information should left to the UI, not the client.  This also makes it easy for the UI to later change how the messages should look, such as using a popup notification.


Dragonisser (approved):
Looks good to me

PR #389, opened 2022-07-23
SkyrimTogether.esp

RobbeBryssinck:
We can now directly add and edit game records. We could also modify quest scripts and such. This effectively enables the community to redesign the game to be made for multiplayer instead of singleplayer. This would not require C++ coders or reverse engineers. Traditional Papyrus modders could do this.


Force67:
Me likey

PR #397, opened 2022-07-23
Actually Fixed Comment

kulcris:
Seems I put the Type instead of the actual variable name.


kulcris:
not sure why this failed to build o.O

yamashi (approved):

PR #416, opened 2022-07-26
Server command for setting ingame time

Sturlen:
An often requested feature.

~~Submitted as draft becuase `aStack.Pop<uint16_t>()` currently throws with `std::bad_any_cast` regardless of which numeric types I use. Everything else works as expected, so any help on this would be appreciated.~~
EDIT:
Solved with the help of cosideci

### DIFF
--- a/Code/server/GameServer.cpp
+++ b/Code/server/GameServer.cpp
@@ -245,7 +245,7 @@ void GameServer::BindServerCommands()
 
     m_commands.RegisterCommand<>("quit", "Stop the server", [&](Console::ArgStack&) { Kill(); });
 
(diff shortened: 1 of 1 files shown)


RobbeBryssinck:
Tested in game?

Sturlen:
> Tested in game?

Tested ingame with clients connected. Client ingame time is set whenever the command is run and the world reacts appropriately e.g. people get up from bed if you set it to midday if it was night previously. Will also sync if you join after then change has happened.

This PR does not change how the CalenderSerivice works, so any previous time or calendar bugs would still be there.

PR #425, opened 2022-07-28
Simplify handling of chat messages between Client and UI

Sturlen:
Removes 'dialogue' and 'system' message events and handlers, and sends all types of chat messages via the 'message' event instead.
Messages now all have a type/opcode instead:
```
export enum MessageType {
  SYSTEM_MESSAGE = 0,
  GLOBAL_CHAT = 1,
  PLAYER_DIALOGUE = 2,
}
```
This makes it much easier to add new message types such as whispers, party, or local chat, since you don't need to create an event and an event handler for each one.
This change is entirely for improving the internals and will have no visible effects for players.

I'm aware this is quite a large change, but I think it's a necessary one for futureproofing.

Tested in game. System, chat and dialogue messages both transmit and display correctly.

EDIT: 
System message
system
Chat
chat
Dialogue
dialogue


Code/skyrim_ui/src/app/components/chat/chat.component.ts
```diff
@@ -2,14 +2,28 @@ import { AfterViewChecked, Component, ElementRef, HostListener, QueryList, ViewC
...
 import { Sound, SoundService } from '../../services/sound.service';
 import { MessageHistory } from './message-history';
 
 interface ChatMessage extends Message {
   date: number;
   odd: boolean;
+  typeClass: string;
+}
+
+function messagegeTypeToClassName(type: MessageType): string {
```
scomans, line 16: typo

scomans, line 16: typo

Sturlen: fixed now

Sturlen:
Needs some more work.

PR #427, opened 2022-07-28
Add TimeScale Server Setting

Sturlen:
Let you set how fast time passes on the server. 
- 0 will stop the clock
- 1 will move at realtime
- 1000 set it to 50x skyrim's normal speed

Tested in game with two clients. Left on for an hour at 1000 timescale with no noticable effects, other than water moving animating way too fast.

https://user-images.githubusercontent.com/43607012/181633426-df0830c2-1ecd-42a8-8d2e-427f227efd37.mp4

### DIFF
--- a/Code/server/Services/CalendarService.cpp
+++ b/Code/server/Services/CalendarService.cpp
@@ -84,3 +84,19 @@ CalendarService::TDate CalendarService::GetDate() const noexcept
 {
     return {m_timeModel.Day, m_timeModel.Month, m_timeModel.Year};
 }
+
+bool CalendarService::SetTimeScale(float aScale) noexcept
+{
+    if (aScale >= 0.f && aScale <= 1000.f)
+    {
+        m_timeModel.TimeScale = aScale;
+
+        ServerTimeSettings timeMsg;
+        timeMsg.TimeScale = m_timeModel.TimeScale;
+        timeMsg.Time = m_timeModel.Time;
+        GameServer::Get()->SendToLoaded(timeMsg);
+        return true;
+    }
+
+    return false;
+}
--- a/Code/server/Services/CalendarService.cpp
+++ b/Code/server/Services/CalendarService.cpp
@@ -94,7 +94,7 @@ bool CalendarService::SetTimeScale(float aScale) noexcept
         ServerTimeSettings timeMsg;
         timeMsg.TimeScale = m_timeModel.TimeScale;
         timeMsg.Time = m_timeModel.Time;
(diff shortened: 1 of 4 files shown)


invaderdimz:
Make sure this is even viable because Time Scale might mess up some quests. Especially ones that go off time. I'm assuming this because in Fallout New Vegas ik adjusting time scale will break some quests.

RobbeBryssinck:
I don't see the use of this.

Sturlen:
> Make sure this is even viable because changing Time Scale might mess up some quests.

I don't recall any quests with timelimits, but it's been a while since I did a full playthrough. If you know of any, let me know and I can test them.

> I don't see the use of this.

It's been requested on the discord quite a lot:
"I wish there was a way to modify the timescale though. You'll never experience half the seasons at the rate it goes"
"Is it possible to tweak the server timescale?"
"how do I change the timescale?"
"can you change the date and timescale when hosting a server?"
and so forth.

 Some people want to see the seasons pass them by and some want time to feel real.

RobbeBryssinck:
This will most likely have unforeseen gameplay implications. Also, skyrim doesn't have a concept of seasons.

Sturlen:
> This will most likely have unforeseen gameplay implications.

I do agree that it's currently untested territory and I could add a warning if you want. Still, setting TimeScale on clients is an existing feature and this just lets users access it. If the setting the TimeScale is too unstable to ever be used, then why is it fully implemented and synchronized?

> Also, skyrim doesn't have a concept of seasons.

I should have worded it better.
Was meant as the months of the calandar changing, people want it for immersion reasons.

Code/server/Services/CalendarService.cpp
```diff
@@ -84,3 +84,19 @@ CalendarService::TDate CalendarService::GetDate() const noexcept
...
+bool CalendarService::SetTimeScale(float aScale) noexcept
+{
+    if (aScale >= 0.f && aScale <= 1000.f)
+    {
+        m_timeModel.TimeScale = aScale;
+
+        ServerTimeSettings timeMsg;
+        timeMsg.TimeScale = m_timeModel.TimeScale;
+        timeMsg.Time = m_timeModel.Time;
+        GameServer::Get()->SendToLoaded(timeMsg);
```
RobbeBryssinck, line 97: Use SendToPlayers()

RobbeBryssinck (changes requested):
Should make it a server setting then too, not just a command.

Code/server/Services/CalendarService.cpp
Sturlen: fixed

PR #428, opened 2022-07-29
Fix: Limit length of servername in serverlist UI

Sturlen:
^
Limit set to 100 characters, since any lower would break a lot of discord links.

### DIFF
--- a/Code/skyrim_ui/src/app/services/server-list.service.ts
+++ b/Code/skyrim_ui/src/app/services/server-list.service.ts
@@ -6,6 +6,12 @@ import { environment } from '../../environments/environment';
 import { GeoLocation } from '../models/geo-location';
 import { Server } from '../models/server';
 
+const MAX_SERVERNAME_LENGTH = 100;
+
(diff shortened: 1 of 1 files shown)


Sturlen:
Reopened due to discussion in the Discord server.

Code/skyrim_ui/src/app/services/server-list.service.ts
```diff
@@ -20,7 +26,8 @@ export class ServerListService {
   public getServerList(): Observable<Server[]> {
     return this.http.get<Server[]>(`${ environment.urlProtocol }://${ environment.url }/list`)
       .pipe(
-        map((data: any) => data.servers),
+        map<any, Server[]>((data: any) => {console.log(data); return data.servers}),
```
RobbeBryssinck, line 29: console.log seems unnecessary.

Sturlen: I agree, excess logging has now been removed

PR #429, opened 2022-07-29
Truncate Server name when it's loaded from config

Sturlen:
This prevents users from setting server names that are extremely long and announcing them to the Serverlist. 

Limit is set to 100 bytes. Server names over this will result in the name being truncated and an error message in the console.

Tested on STServer and FTServer.

### DIFF
--- a/Code/server/GameServer.cpp
+++ b/Code/server/GameServer.cpp
@@ -20,6 +20,7 @@
 #include <Messages/NotifyPlayerJoined.h>
 #include <console/ConsoleRegistry.h>
 
+constexpr size_t kMaxSererNameLength = 128u;
 
 // -- Cvars --
 Console::Setting<uint16_t> uServerPort{"GameServer:uPort", "Which port to host the server on", 10578u};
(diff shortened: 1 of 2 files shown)


Code/server/GameServer.cpp
```diff
@@ -264,11 +264,21 @@ void GameServer::BindServerCommands()
...
-
+    /* Update Info fields from user facing CVARS.*/
 void GameServer::UpdateInfo()
 {
-    // Update Info fields from user facing CVARS.
-    m_info.name = sServerName.c_str();
+    const String cServerName = sServerName.c_str();
+
+    if (cServerName.length() > cm_MaxServerNameLength) 
+    {
```
Force67, line 273: Make cm_maxservernamelength a constexpr size_t global within the source like constexpr size_t kMaxSererNameLength… also make the limit a power of two.

Force67 (changes requested):

Code/server/GameServer.cpp
Sturlen: fixed

PR #438, opened 2022-07-31
Fix: set maxstdio to a higher value

RobbeBryssinck:
This allows for more mods without save corruption.

### DIFF
--- a/Code/immersive_launcher/Launcher.cpp
+++ b/Code/immersive_launcher/Launcher.cpp
@@ -47,6 +47,21 @@ LaunchContext* GetLaunchContext()
         return false;                                                                                                  \
     }
 
+
+void SetMaxstdio()
+{
+    const auto handle = GetModuleHandleW(L"API-MS-WIN-CRT-STDIO-L1-1-0.DLL");
+    if (!handle)
+        return;
+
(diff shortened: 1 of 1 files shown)


Code/immersive_launcher/Launcher.cpp
```diff
@@ -47,6 +47,21 @@ LaunchContext* GetLaunchContext()
         return false;                                                                                                  \
     }
 
+
+void SetMaxstdio()
+{
+    const auto handle = GetModuleHandleW(L"API-MS-WIN-CRT-STDIO-L1-1-0.DLL");
```
yamashi, line 53: maybe Load instead?

RobbeBryssinck: as in LoadMaxstdio()?

yamashi: LoadModule lol

RobbeBryssinck: oh lol, well, yeah i guess, but the dll is already in memory, so i figured GetModuleHandle would be more efficient

yamashi: it's loaded on your PC, might not be on other machines

Force67: You are referencing a shim dll of ucrtbase.dll, could use that directly.

PR #439, opened 2022-07-31
UI: mock skyrimtogether instead of faking it via a service & E2E

scomans:
Allows easier and more robust debugging of the UI

Feedback is welcome 😀


Sturlen:
The new debug menu feels lovely to use and it's very nice to see the all mocking code in one place.

Would it be possible to use the drag feature on other components such as the GroupComponent? Feels a lot better to just drag stuff where you want it, rather than setting x and y positions through a menu.

scomans:
I don't think it's a good idea to change any components used ingame to use drag with the current CEF performance 😅

RobbeBryssinck:
Fantastic work, should make future debugging much easier! Has it been tested that this does not break anything in game?

scomans:
I tried it ingame and it works fine.

PR #441, opened 2022-08-02
Fix Player Summons

### LINKED ISSUE #294: Conjuration summons are always flagged as hostile
## Description

A Conjuration summon that was casted by a player in a group would be seen as hostile by other party members or passive/friendly NPCS. If a player Summons an allied entity, that allied entity is always known as hostile from another players perspective in the group. All of the NPCS in a town would start attacking a players summon because the summoned entity is flagged as hostile all the time for no reason.

## Steps to reproduce

1. Start the game
2. Start new game
3. Finish Helgen sequence 
4. Join a party CO-OP
5. Obtain spellbook of any conjuration summon 
6. Be in a city with CO-OP group 
7. Cast conjuration summon 
8. All party members will see casters summon as hostile 
9. All NPCS will now randomly start attacking casters summon randomly 

## Reproduction rate

Please check just _one_ of the following reproduction rates:

- [*] Consistent: I can reproduce this bug by following the steps to reproduce 100% of the time (*every time*)
- [ ] Occasional: I can reproduce this bug by following the steps to reproduce around 50% of the time
- [ ] Rare: I can reproduce this bug by following the steps to reproduce around 10% or less of the time

## Expected result

I expected a players casted conjuration summon to be friendly to the party group and passive/friendly NPCS.

Sturlen:
This makes player summons friendly to all players, as if they were your own. This also makes them stop fighting eachoter. This works by setting CommandingActor on remote summons to the local player Actor.

This early draft features a lot of workarounds and is meant as a proof of concept for testing and further refining.

Requires each player to first summon a creature, in order to obtain their `ActorHandle` before other player summons can become friendly. This could likely be improved by adding GetActorHandle from CommonLib.

This has only been tested and working on **Conjured Player Summons**, not NPC summons or raised corpses. NPC's summon require getting access to their `ActorHandle` and necromancy would need improvment in syncing corpses.

EDIT:
Fixes #294


Code/client/Services/Debug/Views/EntitiesView.cpp
```diff
@@ -170,6 +170,16 @@ void DebugService::DisplayFormComponent(FormIdComponent& aFormComponent) const n
     ImGui::InputScalar("Is dead?", ImGuiDataType_U8, &isDead, 0, 0, "%" PRIx8, ImGuiInputTextFlags_ReadOnly);
     int isRemote = int(pActor->GetExtension()->IsRemote());
     ImGui::InputScalar("Is remote?", ImGuiDataType_U8, &isRemote, 0, 0, "%" PRIx8, ImGuiInputTextFlags_ReadOnly);
+    auto owner = pActor->GetCommandingActor();
+    int commandingActorId = int(owner ? owner->formID : 0x0);
+    if (ImGui::Button("Take Commmand"))
+    {
+        pActor->SetCommandingActor(m_world.playerHandler);
```
RobbeBryssinck, line 177: Don't call in game functions that change state, like setters, from the ImGui thread. It's not thread safe. I understand that this is test code and will be deleted, but just letting you know for the future.

Code/client/Services/Generic/CharacterService.cpp
```diff
@@ -173,12 +174,31 @@ void CharacterService::DeleteTempActor(const uint32_t aFormId) noexcept
...
     }
 
+    auto commandingActor = pActor->GetCommandingActor();
+    if (commandingActor)
+    {
+        auto pActorHandle = pActor->currentProcess->middleProcess->commandingActor;
+        auto pOwner = TESObjectREFR::GetByHandle(pActorHandle.handle.iBits);
+        auto pOwnerActor = Cast<Actor>(pOwner);
+
+        if (pOwner && pOwnerActor && pOwnerActor->formID == 0x14)
```
RobbeBryssinck, line 191: pOwner check is not necessary

Code/client/World.h
```diff
@@ -45,6 +46,8 @@ struct World : entt::registry
     static void Create() noexcept;
     [[nodiscard]] static World& Get() noexcept;
 
+    ActorHandle playerHandler;
```
RobbeBryssinck, line 49: This shouldn't be stored in World. Use Actor::GetHandleByFormId() or whatever the in game function is.

Code/client/Services/Generic/CharacterService.cpp
```diff
@@ -493,6 +513,14 @@ void CharacterService::OnCharacterSpawn(const CharacterSpawnRequest& acMessage)
     if (pActor->IsDead() != acMessage.IsDead)
         acMessage.IsDead ? pActor->Kill() : pActor->Respawn();
 
+    spdlog::info("Spawn Reqeust Is summon {} {}", acMessage.IsPlayerSummon, m_world.playerHandler ? "handle" : "no handle");
+
+    if (acMessage.IsPlayerSummon && m_world.playerHandler)
```
RobbeBryssinck, line 518: This is flawed. If the local player never summoned anything, there'd be no "playerHandler".

RobbeBryssinck (changes requested):
Don't forget to remove your logs. Also, to what extent has this been tested? What happens when the remote player tries to summon their own actor? What happens when the owning player fast travels away?

Code/client/Services/Debug/Views/EntitiesView.cpp
Sturlen: good to know, thanks!

RobbeBryssinck:
Keep in mind that fallout together has to build too now.

Code/client/World.h
Sturlen: Using this workaround because I haven't been able to implement Actor::GetHandle() yet. Will be removed when I do.

Code/client/Services/Generic/CharacterService.cpp
```diff
@@ -493,6 +513,14 @@ void CharacterService::OnCharacterSpawn(const CharacterSpawnRequest& acMessage)
...
+    {
+        pActor->SetCommandingActor(m_world.playerHandler);
+        m_world.GetRunner().Trigger(MoveActorEvent(pActor->formID, pActor->GetParentCell()->formID, pActor->position));
```
RobbeBryssinck, line 521: Why move it to itself?

Sturlen: The Actor won't actually become a real CommandedActor until you trigger the MoveActorEvent, unsure why. I'm guessing it has something to do with updating the actor's state. Will be removed onces I've found a better method.

RobbeBryssinck:
In that case, it does take the original summon with it, which makes sense, but your server code is faulty then.

Code/client/Components/PlayerSummonComponent.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
```
RobbeBryssinck, line 1: Why PlayerSummonComponent?

Sturlen: I expected I would need to store more data to know who the owner is. Turned out to be kinda unnecessary. I intend to refactor this into something simpler

RobbeBryssinck: Ah, I've been refactoring it a bit already, I'll push it in a bit.

Code/client/Services/Debug/Views/EntitiesView.cpp
Sturlen: has now been removed

PR #451, opened 2022-08-08
Add missing German and Dutch translation strings for party invite

Sturlen:
Just noticed SERVICE.PLAYER_LIST.PARTY_INVITE was missing in both languages.
Great if @RobbeBryssinck and @scomans could check the NL and DE translations respectively.

### DIFF
--- a/Code/skyrim_ui/src/assets/i18n/de.json
+++ b/Code/skyrim_ui/src/assets/i18n/de.json
@@ -138,6 +138,9 @@
         "CLIENT_MODS_DISALLOWED": "Dieser Server verbietet folgende Mods {{mods}}. Bitte entferne diese, um beitreten zu können.",
         "WRONG_PASSWORD": "Das eingegebene Passwort ist falsch.",
         "NO_REASON": "Der Server hat die Verbindung ohne Angabe von Gründen verweigert."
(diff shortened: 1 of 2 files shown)


Force67:
german part LGTM

NL-TCH (approved):

Code/skyrim_ui/src/assets/i18n/de.json
```diff
@@ -139,6 +139,9 @@
         "WRONG_PASSWORD": "Das eingegebene Passwort ist falsch.",
         "NO_REASON": "Der Server hat die Verbindung ohne Angabe von Gründen verweigert."
       }
+    },
+    "PLAYER_LIST": {
+      "PARTY_INVITE": "{{from}} hat Sie zu einer Gruppe eingeladen"
```
scomans, line 144: I would use the translation below as it isn't as formal

`{{from}} hat dich in eine Gruppe eingeladen`

scomans (changes requested):

Code/skyrim_ui/src/assets/i18n/de.json
Sturlen: changed

PR #457, opened 2022-08-12
Fix bug in server settings config load/save

Outshynd:
Fix for a bug where uServerPort and uMaxPlayerCount would neither load from or save to config file.

### DIFF
--- a/Code/server/GameServer.cpp
+++ b/Code/server/GameServer.cpp
@@ -23,8 +23,8 @@
 constexpr size_t kMaxSererNameLength = 128u;
 
 // -- Cvars --
-Console::Setting<uint16_t> uServerPort{"GameServer:uPort", "Which port to host the server on", 10578u};
-Console::Setting<uint16_t> uMaxPlayerCount{"GameServer:uMaxPlayerCount", "Maximum number of players allowed on the server (going over the default of 8 is not recommended)", 8u};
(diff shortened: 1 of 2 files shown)


RobbeBryssinck:
Could you elaborate a bit on this? I find it unlikely that setting the port did not work at all before this fix, since it has been fairly heavily tested.

Outshynd:
Sure.  It's pretty simple: in [a recent commit](https://github.com/tiltedphoques/TiltedEvolution/commit/68a6df8af16e806cdc089995aafa8efa296f5ac5), `<uint16_t>` was added to settings uServerPort and uMaxPlayerCount, which broke those two settings from loading correctly or saving.  This PR is simply changing it back to just `Console::Setting` which fixes the issue.  The values are explicitly cast when used later in the code anyway, as with `auto port = uServerPort.value_as<uint16_t>();`.

In fact, if you change just the first setting, removing `<uint16_t>` from uServerPort but leaving it for uMaxPlayerCount, uServerPort is still broken.  So seemingly anything that appears before the incorrect setting breaks.  I didn't do any debugging to find where or why it breaks in that way, but did notice removing the initial type cast does fix the issue, both in linux and Windows.

Outshynd:
It would seem this is now not building on Windows, though I am quite sure I built it on Win10.  Seems there's more to it than just removing `<uint16_t>` from the setting.  Builds fine on Linux strangely.  Gonna convert to a Draft for now.

PR #463, opened 2022-08-16
Tweak: A few minor tweaks / fixes for the server

Outshynd:
Fixed a typo in `kMaxSererNameLength` variable.  Changed new player and player disconnected console messages to include character name.  Added command `/SetDifficulty`.  Changed hard-coded instances of `/` to use `kCommandPrefix` just in case the command prefix is ever changed.

### DIFF
--- a/Code/server/GameServer.cpp
+++ b/Code/server/GameServer.cpp
@@ -23,8 +23,8 @@
 constexpr size_t kMaxSererNameLength = 128u;
 
 // -- Cvars --
-Console::Setting<uint16_t> uServerPort{"GameServer:uPort", "Which port to host the server on", 10578u};
-Console::Setting<uint16_t> uMaxPlayerCount{"GameServer:uMaxPlayerCount", "Maximum number of players allowed on the server (going over the default of 8 is not recommended)", 8u};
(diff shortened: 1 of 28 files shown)


Code/server/GameServer.cpp
```diff
@@ -58,6 +58,22 @@ Console::Command<> TogglePvp("TogglePvp", "Toggle PvP on/off", [](Console::ArgSt
     spdlog::get("ConOut")->info("PvP has been {}.", bEnablePvp == true ? "enabled" : "disabled");
 });
 
+Console::Command<int64_t> SetDifficulty("SetDifficulty", "Set server difficulty (0 being Novice and 5 being Legendary; default is 4)", [](Console::ArgStack& aStack) {
```
RobbeBryssinck, line 61: Currently connected clients need a message telling them to change the difficulty, otherwise the difficulty will be desynced.

RobbeBryssinck (changes requested):

Code/server/GameServer.cpp
Outshynd: Resolved by Sturlen in merged pr below

PR #474, opened 2022-08-23
Player aggro division 

RobbeBryssinck:
This should make it so that enemies targets the entire party more evenly. An enemy will target whichever player has hit them last.


Code/client/Services/CombatService.h
```diff
@@ -0,0 +1,40 @@
...
+struct ProjectileLaunchedEvent;
+struct NotifyProjectileLaunch;
+struct HitEvent;
+
+/**
+* @brief Responsible for projectiles, combat agro, etc.
+*/
+struct CombatService
+{
+public:
```
yamashi, line 15: not needed

Code/client/Services/Generic/CombatService.cpp
```diff
@@ -0,0 +1,277 @@
...
+    const uint32_t cSpellId = modSystem.GetGameId(acMessage.SpellID);
+    launchData.pSpell = Cast<MagicItem>(TESForm::GetById(cSpellId));
+
+    launchData.eCastingSource = (MagicSystem::CastingSource)acMessage.CastingSource;
+
+    launchData.iArea = acMessage.Area;
+    launchData.fPower = acMessage.Power;
+    launchData.fScale = acMessage.Scale;
+
+    launchData.bAlwaysHit = acMessage.AlwaysHit;
```
yamashi, line 164: Could be better to name them the same?

Code/client/Games/Skyrim/Actor.h
```diff
@@ -234,11 +234,18 @@ struct Actor : TESObjectREFR
     void Kill() noexcept;
     void Reset() noexcept;
     void Respawn() noexcept;
+
     void PickUpObject(TESObjectREFR* apObject, int32_t aCount, bool aUnk1, float aUnk2) noexcept;
     void DropObject(TESBoundObject* apObject, ExtraDataList* apExtraData, int32_t aCount, NiPoint3* apLocation, NiPoint3* apRotation) noexcept;
     void DropOrPickUpObject(const Inventory::Entry& arEntry, NiPoint3* apPoint, NiPoint3* apRotate) noexcept;
     void SpeakSound(const char* pFile);
 
+    bool IsInCombat() noexcept;
```
yamashi, line 243: const [[nodiscard]]

Code/client/Games/Skyrim/Actor.h
```diff
@@ -234,11 +234,18 @@ struct Actor : TESObjectREFR
...
+    Actor* GetCombatTarget() noexcept;
```
yamashi, line 244: const [[nodiscard]]

yamashi (changes requested):
Nice job, very minor changes

Code/client/Services/Generic/CombatService.cpp
RobbeBryssinck: I considered that, but the member names in LaunchData are closer to the actual names used in the in-game struct. Could do either way.

kulcris:
I am curious if it will cause target swapping a bit too much though. resulting in playing a game of keep away with a single enemy. 
Could this be adjusted to be the highest damaging player instead?

RobbeBryssinck:
> I am curious if it will cause target swapping a bit too much though. resulting in playing a game of keep away with a single enemy. Could this be adjusted to be the highest damaging player instead?

Yes, and I did design it with that in mind, so that the enemy is locked onto one player for 5 seconds, but I see now that there's a bug and that that timer is ignored. I will fix this tomorrow.

PR #477, opened 2022-08-29
Chat commands for party and local chat

Sturlen:
### Summary
Adds Party chat and Local chat. Party chat is only visible for members of your party and local chat will be seen by players in the same interior or close enough distance outdoors.

Used by typing `/local MESSAGE` or `/party MESSAGE`.

msgtypes

In the future I plan on adding a button/dropdown to let you more easily select a mode.
`/whisper PLAYER` is also in the works.

A `/help` command has also been added for finding all available commands.

### Details
NotifyChatMessageBroadcast has been extended to support multiple type of chat messages, so that each type does not need to create a seperate serialization function and handlers.
```
enum ChatMessageType : unsigned char
{
    kSystemMessage = 0,
    kGlobalChat,
    kPlayerDialogue,
    kPartyChat,
    kLocalChat
};
```
Commands are now only handled by the UI, and uses the existing event bus to the Client. Client-UI events have been updated to pass along a chat message type, both for sent and recieved messages. `/tp` was removed because it is already handled by the UI.

A seperate ChatService has been made which now handles all chat messages and commands. Other services can use `ChatService.register(Command)` to register new commands, which will also show up in the help menu. Commands are passed args that were seperated by a space, therefore any commands relating to players shold pass along their Id instead of their name.

Translations of system messages and notification are now done in the template, which means previous messages will update if you change your language preference.

As always, please let me know if you find any mistakes or bugs I missed.


Code/skyrim_ui/src/app/components/chat/chat.component.html
```diff
@@ -2,13 +2,13 @@
   <p
     #entry
     *ngFor="let message of messages"
-    [attr.data-system]="!message.name"
     [attr.data-odd]="message.odd"
```
scomans, line 5: could you replace the odd attribute of the message with the odd alias of the ngFor

Code/skyrim_ui/src/app/services/chat/commands.ts
```diff
@@ -0,0 +1,63 @@
...
+  }}
+
+  private readonly commands: Map<string, Command>;
+
+  private readonly chatService: ChatService
+
+  public constructor (
+    chatService: ChatService
+  ) {
+    this.chatService = chatService;
```
scomans, line 32:
```suggestion
  public constructor (
    private readonly chatService: ChatService,
  ) {
```

Code/skyrim_ui/src/app/services/chat/commands.ts
```diff
@@ -0,0 +1,63 @@
...
+    const content = args.join(' ');
+    this.chatService.sendMessage(MessageTypes.PARTY_CHAT, content);
+  }}
+
+  private LocalChat: Command = { name: 'local', executor: async (args) => {
+    const content = args.join(' ');
+    this.chatService.sendMessage(MessageTypes.LOCAL_CHAT, content);
+  }}
+
+  private readonly commands: Map<string, Command>;
```
scomans, line 25:
```suggestion
  private readonly commands = new Map<string, Command>();
```

Code/skyrim_ui/src/app/services/chat/commands.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  private PartyChat: Command = { name: 'party', executor: async (args) => {
+    const content = args.join(' ');
+    this.chatService.sendMessage(MessageTypes.PARTY_CHAT, content);
+  }}
+
+  private LocalChat: Command = { name: 'local', executor: async (args) => {
+    const content = args.join(' ');
+    this.chatService.sendMessage(MessageTypes.LOCAL_CHAT, content);
+  }}
```
scomans, line 23: Move them outside the class

Code/skyrim_ui/src/app/services/chat/commands.ts
```diff
@@ -0,0 +1,63 @@
...
+  public readonly COMMAND_PREFIX = '/';
+
+  public register(cmd: Command ) {
+    if (!this.commands.has(cmd.name)) {
+      this.commands.set(cmd.name, cmd);
+    }
+  }
+
+  public async tryExecute(input: string) {
+    if (!input.startsWith(this.COMMAND_PREFIX) || input.length < this.COMMAND_PREFIX.length) {
```
scomans, line 49: i think the first check should suffice. Maybe make them case insensitive too

Code/client/Services/Generic/OverlayService.cpp
```diff
@@ -341,8 +343,10 @@ void OverlayService::OnChatMessageReceived(const NotifyChatMessageBroadcast& acM
         return;
 
     auto pArguments = CefListValue::Create();
-    pArguments->SetString(0, acMessage.PlayerName.c_str());
+    pArguments->SetInt(0, acMessage.MessageType);
```
Force67, line 346: cast to int explicitly or make type int

Code/server/Services/OverlayService.cpp
```diff
@@ -24,29 +26,55 @@ OverlayService::OverlayService(World& aWorld, entt::dispatcher& aDispatcher)
...
-    GameServer::Get()->SendToPlayers(notifyMessage);
-}
+    auto character = aSendingPlayer->GetCharacter();
 
-void OverlayService::OnPlayerDialogue(const PacketEvent<PlayerDialogueRequest>& acMessage) const noexcept
-{
-    auto& message = acMessage.Packet;
+    switch (notifyMessage.MessageType)
+    {
+    case (kGlobalChat):
```
Force67, line 41: no round brackets at cases

Force67 (changes requested):

Code/skyrim_ui/src/app/services/chat/commands.ts
Sturlen: fixed

Code/skyrim_ui/src/app/components/chat/chat.component.html
scomans: Even better

Code/skyrim_ui/src/app/services/chat/commands.ts
Sturlen: fixed

Code/skyrim_ui/src/app/services/chat/commands.ts
scomans: Just mark them as resolved

Code/skyrim_ui/src/app/services/chat/commands.ts
Sturlen: moved the LocalChat and PartyChat into the ChatService, since they're both depending on that. Help command will have to stay, as it requires access to `prviate commands` and does not need to access from anywhere else.

Code/skyrim_ui/src/app/services/chat/commands.ts
Sturlen: agreed. fixed

Code/client/Services/Generic/OverlayService.cpp
Sturlen: cast to int is now explict.

Code/server/Services/OverlayService.cpp
Sturlen: brackets removed

Code/skyrim_ui/src/app/services/chat/commands.ts
scomans: Didn't noticed that 🤔 well it doesn't really matter so just let it be as is

PR #483, opened 2022-09-07
Connect dialog styling fixes

Sturlen:
- Save password is now readable and has proper margins
- Input boxes light up when hovered or focused, so you can see where you are typing even if the caret doesn't load
- Server list no longer skips around when scrolling. This was solved by making the font size fixed again so it doesn't interfere virtual scroll.

Before:
image

After:
image

### DIFF
--- a/Code/skyrim_ui/src/app/components/connect/connect.component.html
+++ b/Code/skyrim_ui/src/app/components/connect/connect.component.html
@@ -1,8 +1,10 @@
-<div class="app-connect">
-  <p [innerHtml]="'COMPONENT.CONNECT.INFO.SERVER_ADDRESS' | transloco"></p>
-  <p style="color:rgb(247, 54, 54);">
-    {{ 'COMPONENT.CONNECT.INFO.HELGEN' | transloco }}
-  </p>
+<div class="container">
+  <div id="instructions">
+    <p [innerHtml]="'COMPONENT.CONNECT.INFO.SERVER_ADDRESS' | transloco"></p>
+    <p class="warning">
+      {{ 'COMPONENT.CONNECT.INFO.HELGEN' | transloco }}
+    </p>
+  </div>
 
   <input
     #input
--- a/Code/skyrim_ui/src/app/components/connect/connect.component.html
+++ b/Code/skyrim_ui/src/app/components/connect/connect.component.html
@@ -1,5 +1,5 @@
 <div class="container">
-  <div id="instructions">
+  <div class="instructions">
     <p [innerHtml]="'COMPONENT.CONNECT.INFO.SERVER_ADDRESS' | transloco"></p>
     <p class="warning">
       {{ 'COMPONENT.CONNECT.INFO.HELGEN' | transloco }}
--- a/Code/skyrim_ui/src/app/components/connect/connect.component.scss
+++ b/Code/skyrim_ui/src/app/components/connect/connect.component.scss
@@ -9,6 +9,12 @@
   display: flex;
   flex-direction: column;
   align-items: center;
+
+  ::ng-deep {
+    strong {
+      color: mod($-color-background, 1);
+    }
+  }
 }
 
 .save-password {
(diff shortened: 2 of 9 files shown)


Code/skyrim_ui/src/app/components/connect/connect.component.html
```diff
@@ -1,8 +1,10 @@
-<div class="app-connect">
-  <p [innerHtml]="'COMPONENT.CONNECT.INFO.SERVER_ADDRESS' | transloco"></p>
-  <p style="color:rgb(247, 54, 54);">
-    {{ 'COMPONENT.CONNECT.INFO.HELGEN' | transloco }}
-  </p>
+<div class="container">
+  <div id="instructions">
+    <p [innerHtml]="'COMPONENT.CONNECT.INFO.SERVER_ADDRESS' | transloco"></p>
```
scomans, line 3: Why innerHtml?

Code/skyrim_ui/src/app/components/connect/connect.component.html
```diff
@@ -1,8 +1,10 @@
-<div class="app-connect">
-  <p [innerHtml]="'COMPONENT.CONNECT.INFO.SERVER_ADDRESS' | transloco"></p>
-  <p style="color:rgb(247, 54, 54);">
-    {{ 'COMPONENT.CONNECT.INFO.HELGEN' | transloco }}
-  </p>
+<div class="container">
+  <div id="instructions">
```
scomans, line 2: Use class instead of id

Code/skyrim_ui/src/app/components/connect/connect.component.scss
```diff
@@ -3,56 +3,47 @@
...
+  text-align: center;
+  margin: 0.3em;
+  line-height: $-size-line;
+}
 
-  ::ng-deep {
-    strong {
-      color: mod($-color-background, 1);
-    }
+::ng-deep {
```
scomans, line 28: Put it inside of a class. Otherwise it can pollute other components

Code/skyrim_ui/src/app/components/server-list/server-list.component.scss
```diff
@@ -8,7 +8,7 @@
   display: flex;
   gap: 1em;
   width: 70vw;
-  --server-list-row-height: calc(16 * 100vw / 1920 * 2);
+  --server-list-row-height: 40px;
```
scomans, line 11: This will make the rows very tiny for 4k+ screens.

Code/skyrim_ui/src/app/components/server-list/server-list.component.ts
```diff
@@ -138,6 +141,11 @@ export class ServerListComponent {
       favoriteServers[`${ favoriteServer.ip }:${ favoriteServer.port }`] = favoriteServer;
     }
     this.favoriteServers.next(favoriteServers);
+
+    settingService.fontSizeChange.subscribe(fontSize => {
```
scomans, line 145: manage subscription `.pipe(takeUntil(this.destroy$))`

Code/skyrim_ui/src/app/components/server-list/server-list.component.ts
```diff
@@ -48,14 +50,15 @@ export class ServerListComponent {
   clientVersion$: Observable<string>;
 
   formSearch = new FormControl<string>('');
-  rowHeight = 16 * document.documentElement.clientWidth / 1920 * 2;
+  rowHeight = new BehaviorSubject(16)
```
scomans, line 53: replace `16` with a dynamic value or replace the `BehaviorSubject` with a `ReplaySubject`

Code/skyrim_ui/src/app/components/server-list/server-list.component.ts
Sturlen: I've switched it to using an observable instead now. This is used by an [AsyncPipe ](https://angular.io/api/common/AsyncPipe#description) in the template, which handles unsubbing when the component is destroyed.

PR #484, opened 2022-09-07
WIP: Easy build bat

GokuWorks:
Create bat file to easily build release.


GokuWorks:
I think I fixed the UI env but feel free to let me know if its still wrong.

Easy Build Release.bat
```diff
@@ -0,0 +1,10 @@
+git status
```
RobbeBryssinck, line 1: Not needed

Easy Build Release.bat
```diff
@@ -0,0 +1,10 @@
...
+git fetch
```
RobbeBryssinck, line 2: This does not do anything on its own except fetch metadata. These git operations can be done by the programmer.

Easy Build Release.bat
```diff
@@ -0,0 +1,10 @@
...
+git submodule update --recursive
```
RobbeBryssinck, line 3: Likewise, submodules should not always be updated automatically, this is a conscious choice. Updating submodules should not be in the build script.

RobbeBryssinck (changes requested):
Filename should be "Build.bat".

PR #495, opened 2022-09-13
Tweak for multi-stage, multi-arch build for docker server

Outshynd:
In an effort to teach myself a bit more about Docker, I set about reconfiguring the docker build process for the server.

I matched the docker build process a bit more closely to the `Build linux` workflow, just in case there are any build differences that don't catch edge-case issues.  This includes using `ubuntu:22.04` as a base, installing gcc-12/g++-12 with apt and using `apt update-alternatives` to set it as default.

I also used a little bit of intermediate build trickery to copy only the architecture specific libstdc++.so.6, though I placed it in the server directory (instead of system lib directory) to get around platform issues.

Added `docker-multiarch-build.sh` which automates the build process, including pushing to docker hub.  The commands for building manually are contained in this file.

I changed docker-compose.yml to be more in line with pulling and running a server, enabling a few tweaks like mounting localtime for correct timestamped logs and enabing the equivalent of `-it` flags to allow users to attach and send commands.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,29 +8,28 @@ WORKDIR /home/builder
 ENV XMAKE_ROOT=y
 
 RUN git clone --recursive -b ${BRANCH} ${REPO} ./str && \
-    cd str && xmake config -m release -y && xmake -y && xmake install -o package -y && \
-    cp /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30 /home/builder/libstdc++.so.6 2>/dev/null || :
+    cd str && xmake config -m release -y && xmake -y && xmake install -o package -y
 
 
-#Building for x86_64
+# Building for x86_64
 FROM builder as amd64builder
 
 RUN cp /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30 /home/builder/libstdc++.so.6 2>/dev/null || :
 
 
-#Building for arm64/v8
+# Building for arm64/v8
 FROM builder as arm64builder
 
 RUN cp /usr/lib/aarch64-linux-gnu/libstdc++.so.6.0.30 /home/builder/libstdc++.so.6 2>/dev/null || :
 
 
-#Intermediate image that has the library specific to our $TARGETARCH
+# Intermediate image that has the library specific to our $TARGETARCH
 FROM ${TARGETARCH}builder as intermediate
+# If a user has built without buildx, attempt to save them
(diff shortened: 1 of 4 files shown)


Outshynd:
I forgot to mention a change over the way the current docker build process is implemented: instead of copying local code and then building, it clones a repo/branch.  This is configurable by either editing `Dockerfile` or by passing `--build-arg REPO=<repo>` and `--build-arg BRANCH=<branch>`, and is defaulted to the master branch.

I made this change to make it fewer steps for someone who simply wants to build the docker image on a test server/architecture/platform/etc.  Fewer commands to type into the terminal.  This does have the drawback of being slightly more difficult to rapidly test local changes (need to push to repo before Docker build) and also I suppose can no longer be built without Internet access, but I feel the pros outweigh the cons.

yamashi:
Instructions need to be added to the build guide before this can be merged

Outshynd:
> Instructions need to be added to the build guide before this can be merged

Sure, I can do that a bit later. Since docker instructions didn't exist originally in the build guide I assumed that wasn't necessary. I'll add it to my existing PR for the wiki in a few hours.

Outshynd:
I'm glad you had me look at this.  I was originally content with having to use `docker buildx build .` to build the image, but it bothered me that the default command that most people will already know, `docker build .`, wouldn't work. I managed to change a few things to allow for both, at least for linux/amd64 and linux/arm64.

I [updated my PR for the gitbook-wiki](https://github.com/tiltedphoques/gitbook-wiki/pull/21) to add an advanced 'Build Docker server image' entry, just above 'Server configuration parameters'. You can preview a live version [here](https://unknown-3.gitbook.io/untitled/guides/server-guide/linux-setup/build-docker-image).

Tested on bare metal Ubuntu 22.04 x86_64, WSL Ubuntu 22.04 x86_64, Oracle Cloud arm-based VPS (can only build arm version, rough equivalent to raspi4b).

yamashi:
Having to edit the Dockerfile is breaking for us, as git will not generate the correct build tags if there are modified files in the repo.

I would also remove option 3 considering the files may change at any time.

Outshynd:
> Having to edit the Dockerfile is breaking for us, as git will not generate the correct build tags if there are modified files in the repo.
> 
> I would also remove option 3 considering the files may change at any time.

Having to edit the Dockerfile is only every necessary for someone who does not control the default docker hub repo. Tiltedphoques would not have to edit anything.

Option #2 and #3 are removed. Gitbook PR is fixed.

PR #501, opened 2022-09-15
Calendar Sync

Sturlen:
# Intent #
The plan is to have the first client to join a server send it's TimeModel ( year, mth, day, hour, minute ) and set that as the starting point for the server to simulate time from. When other clients connect or time is changed via SetTime, the current server TimeModel will be sent to them.

There is the possibilty that changing around the dates can cause issues for some time sensitive quests, but this has not been confirmed yet. If this change proves to be unstable, feel free to close it.

# Current state #
TimeModel is currently just sent with ServerTimeSettings messages
Have been able to overwrite the client yr, mth and date values, after I discovered they are stored as floats.
Have attempted to make TimeModel serializable. Does not currently work, but I'm not sure what it's missing.

Feedback and help is appreciated.


Dragonisser:
Stale PR

PR #505, opened 2022-09-18
fix: respawn crash

stenliis:
reproduction:
-load game
-coc whiterun 
-provoke nearest guard and get killed
-game crash during respawn

fix tested on dev branch + fix: animations with new skyrim update

fixed issue for me, but I am not sure that my change doesn't break anything else.


Force67:
Thanks!

PR #508, opened 2022-09-20
fix: serverlist related crashes

RobbeBryssinck:
Exception shouldn't really be thrown in the event of too many requests to the client.


yamashi:
This needs to be reverted, anyone can forge a packet to crash servers now

PR #510, opened 2022-09-20
SetTime chat command for private servers

RobbeBryssinck:
Could still use some error messages for when the user inputs invalid parameters.


yamashi:
Hours and minutes should be uint8 and use write/read bits for serialization

Sturlen (changes requested):

Dragonisser:
Has been implemented via console commands and overhauled in 1.6. It could be still added as a chat command though there is barely any reason to do so after the time and calendar sync.

Issue #512, opened 2022-09-21
Settings UI is slightly broken


stenliis:
## Version of Skyrim Together
v1.3.2 - official build

## Description

Settings UI is slightly broken.

Changing language don't immediately translate the content of the app-dropdown as can be seen in the first screenshot.
Translation is properly applied after closing and reopening settings.

Multiple app-dropdown can be opened at the same time and it look like a mess as can be seen in the second screenshot.
Easiest way achieves this broken behavior is opening them in order: Position, Font Size, Language. 

## Reproduction rate

100%
![ui-bug](https://user-images.githubusercontent.com/13061508/191583535-168c329b-d4f1-46ed-b4a3-6ba1f2b2b11f.png)
![ui-bug2](https://user-images.githubusercontent.com/13061508/191583561-a568bc2e-6155-48c7-a1ed-26d41fbda708.png)

PR #515, opened 2022-09-25
Bug/dropdowns dont close

tomfull123:
This PR fixes the second issue in this report https://github.com/tiltedphoques/TiltedEvolution/issues/512


Sturlen:
My PR #516 already includes the dropdown fix, I wasn't aware anyone else was working on the UI currently. Your solution is a good one and it seems we reached pretty similar conclusions.
The spelling changes look good, but I think the imports should be reverted to the original style of spaces between the brackets, e.g. ; `import { TranslocoService } from '@ngneat/transloco';`

tomfull123:
Ah ok, no problem, I didn't know anyone else was looking into those issues either, I'll close this PR since I don't think there's much point in making a PR just for spelling mistakes haha

PR #516, opened 2022-09-25
Improved Settings UI

### LINKED ISSUE #512: Settings UI is slightly broken

Sturlen:
Fixes and improvements for the styling and scripting of the Settings UI window. I know this is a large PR, but I think it's best that the changes are reviewed with all the surrounding context. No C++ changes.
# Visuals
image
* The menu has been made wider to better support large font sizes and more verbose languages, which were previously cut off. 
* Input components styling and sounds have are now more consistent. 
* Reworked the dropdown to look more skyrim-like and fixes a numbe of issues such as labels not updating when changing languages. Fixes #512.
* Better skyrim-like styling of scrollbars and sliders.

# Scripting
Settings are now BehaviorSubjects, which are gathered in one settings object, which is accessed from the SettingsService This removes a lot of duplicate handling of reading and updating settings, as well as making sure settings always have a value. Also makes it very easy to add new ones. Added new functions for accessing localstore in a more type-safe manner.

Comments and change request are always welcome.


Code/skyrim_ui/src/app/services/setting.service.ts
```diff
@@ -20,115 +20,87 @@ export const fontSizeToPixels: Record<FontSize, number> = {
...
-    return this.translocoService.getDefaultLang();
+  private readonly languageValues: string[] = Object.values(this.translocoService.getAvailableLangs()).map(lang => lang.id);
+  private readonly fontSizeValues = Object.values(FontSize);
+  private readonly partyAnchorValues = Object.values(PartyAnchor);
+  private readonly autoHideTimeValues = autoHideTimerLengths;
+
+  public settings = {
+    volume: new SliderSetting(this.storeService, "audio_volume", 0.5),
+    muted: new ToggleSetting(this.storeService, "audio_muted", false),
+    language: new SelectSetting(this.storeService, "language", this.languageValues, "en"),
```
tomfull123, line 89: We should probably call `this.translocoService.getDefaultLang()` to get the default, thoughts?

Code/skyrim_ui/src/app/components/settings/settings.component.ts
```diff
@@ -1,25 +1,24 @@
...
 import { ClientService } from 'src/app/services/client.service';
 import { FontSize, SettingService } from 'src/app/services/setting.service';
 import { Sound, SoundService } from '../../services/sound.service';
 
 export enum PartyAnchor {
-  TOP_LEFT,
-  TOP_RIGHT,
-  BOTTOM_RIGHT,
-  BOTTOM_LEFT,
+  TOP_LEFT = 'top_left',
```
tomfull123, line 8: While I do approve of giving these constant values, are we ok with breaking people's existing settings for the party anchor?

Code/skyrim_ui/src/app/components/group/group.component.ts
```diff
@@ -125,24 +125,54 @@ export class GroupComponent implements OnInit, OnDestroy {
...
+  }
+
   ngOnDestroy() {
     if (this.timerSubscription) {
       this.timerSubscription.unsubscribe();
     }
   }
 
   private flashGroup() {
+    const timerLength = parseInt(this.settings.autoHideTime.value) * 1000;
```
tomfull123, line 174: As discussed, this variable would be better suited being closer to where its used.

Code/skyrim_ui/src/app/components/settings/settings.component.ts
```diff
@@ -1,25 +1,24 @@
...
+  TOP_RIGHT = 'top_right',
+  BOTTOM_RIGHT = 'bottom_right',
+  BOTTOM_LEFT = 'bottom_left',
 }
 
+export const autoHideTimerLengths = [1, 3, 5].map(l => l.toFixed(0));
```
tomfull123, line 14: I think that this should be an array of numbers, not strings and that the SelectSettings' options parameter should be an `any[]` or a `(string | number)[]`

Code/skyrim_ui/src/app/components/dropdown/dropdown.component.ts
```diff
@@ -105,15 +100,20 @@ export class DropdownComponent implements AfterViewInit, ControlValueAccessor {
...
-  @HostListener('document:click')
-  onClick() {
+  @HostListener('document:click', ['$event'])
+  onClick(e: PointerEvent) {
+    const target = (e.target as Element);
+    const thisElement = this.elRef.nativeElement as Element;
+    if (thisElement.contains(target)) {
+      return;
+    }
+    this.isOpen && this.soundService.play(Sound.Cancel);
```
tomfull123, line 110: I feel like this would be more readable as an if statement.

Code/skyrim_ui/src/app/components/dropdown/dropdown.component.html
```diff
@@ -7,8 +7,7 @@
   <div class="box" (click)="toggle($event)">
     <ng-container>
       <div class="dropdown-selected">
-        <span>{{ options?.[selected] ? options[selected].text : placeholder }}</span>
-        <fa-icon [icon]="faCaretDown"></fa-icon>
+        <span>{{ (options?.[selected] ? options[selected].text : placeholder) | transloco: {length: options?.[selected]?.value} }}</span>
```
tomfull123, line 10: Because we know pipe the options and the selected value through the transloco function, we now get a bunch of errors for the language dropdown because they don't have translations are already in their respective languages.

tomfull123 (changes requested):

Code/skyrim_ui/src/app/components/settings/settings.component.ts
scomans: Also please make them all upper case.

tomfull123: Also I think this enum definition should be in the setting service, with the FontSize enum.

Code/skyrim_ui/src/app/components/settings/settings.component.html
```diff
@@ -1,144 +1,118 @@
...
-              type="radio"
-              [value]="5"
-              [disabled]="!autoHideParty"
-              [ngModel]="autoHideTime"
-              (ngModelChange)="onAutoHideTimeChange($event)"
-            >
-          </div>
-        </div>
+        <app-dropdown [disabled]="!(settings.autoHideParty | async)" [ngModel]="settings.autoHideTime | async" (ngModelChange)="settings.autoHideTime.next($event)">
+          <app-dropdown-option *ngFor="let t of availableAutoHideTimes" [text]="'COMPONENT.SETTINGS.LENGTH'" [value]="t"></app-dropdown-option>
```
scomans, line 68: please don't use just `t` as variable name

scomans (changes requested):
Where did you get the checkbox images from?

Code/skyrim_ui/src/app/components/group/group.component.ts
```diff
@@ -125,24 +125,54 @@ export class GroupComponent implements OnInit, OnDestroy {
   }
 
   private onPartyAutoHideState() {
-    this.settings.partyAutoHideChange
+    this.settings.autoHideParty
       .pipe(takeUntil(this.destroy$))
       .subscribe((state: boolean) => {
         this.isAutoHide.next(state);
-        if (this.settings.isPartyShown()) {
+        if (this.settings.isPartyShown.value) {
```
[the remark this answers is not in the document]
tomfull123: I think the `value` getter is nicer than using the `getValue()` function, unless there's another reason to do so that I'm unaware of?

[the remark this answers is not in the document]
scomans: well `value` internally calls `getValue()` so why don't use it directly?

Sturlen:
> Where did you get the checkbox images from?

They are still the same ones added in #318 , I've just modified the colors so that they can be fully white when hovered.

Code/skyrim_ui/src/app/components/group/group.component.ts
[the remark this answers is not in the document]
tomfull123: It's splitting hairs, but personally I find `value` slightly more readable and its slightly less verbose

Code/skyrim_ui/src/app/components/settings/settings.component.ts
```diff
@@ -78,12 +74,6 @@ export class SettingsComponent implements OnInit {
     this.language = this.settingService.getLanguage();
   }
 
-  onShowDebugChange(checked: boolean) {
-    this.settingService.setDebugShown(checked);
-    this.client.debugStateChange.next(checked);
```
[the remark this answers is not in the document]
Sturlen: There are no listeners for ClientService.debugStateChange, so I don't see why it can't be removed. It seems to be overwriting a variable sent by the client, which is never used and changing it in the UI code will not send the changes back to the Client.

Code/skyrim_ui/src/app/components/group/group.component.ts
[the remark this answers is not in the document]
Sturlen: I agree that `value` is more readable, but the rest of the codebase uses `getValue()` so I've gone with that for consistency. 
If the standard of the codebase should be changed to `value`, then that would be a better fit for separate PR.

Code/skyrim_ui/src/app/components/settings/settings.component.ts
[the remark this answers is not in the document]
tomfull123: Cool, that's fine then, figured it was worth checking.

Code/skyrim_ui/src/app/services/setting.service.ts
```diff
@@ -20,115 +20,87 @@ export const fontSizeToPixels: Record<FontSize, number> = {
...
-  public getAutoHideTime(): number {
-    return JSON.parse(this.storeService.get('party_autoHideTime', 3));
+  constructor(private readonly storeService: StoreService, private storeKey: string, defaultValue: number) {
+    const inital_value = storeService.getFloat(storeKey, defaultValue);
+    super(inital_value)
   }
 
-  public setPartyAnchor(anchor: PartyAnchor) {
-    this.storeService.set('party_anchor', anchor);
+  public next(value: number) {
```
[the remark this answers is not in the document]
Sturlen: I'd rather leave the specifics of how to store and retrieve settings values to the child classses. This way lets you implement settings which don't use localstorge for persistence later down the line, for example reading from an INI file.

Code/skyrim_ui/src/app/components/dropdown/dropdown.component.html
```diff
@@ -7,8 +7,7 @@
   <div class="box" (click)="toggle($event)">
     <ng-container>
       <div class="dropdown-selected">
-        <span>{{ options?.[selected] ? options[selected].text : placeholder }}</span>
-        <fa-icon [icon]="faCaretDown"></fa-icon>
+        <span>{{ noTranslate ? getSelectedLabel() : (getSelectedLabel() | transloco: {length: options[selected]?.value}) }}</span>
```
scomans, line 10:
```suggestion
        <span>{{ noTranslate ? selectedLabel : (selectedLabel | transloco: { length: options[selected]?.value }) }}</span>
```
Could you convert the function to a getter. Using functions in template is considered non optimal practice

Code/skyrim_ui/src/app/components/dropdown/dropdown.component.ts
```diff
@@ -80,22 +81,28 @@ export class DropdownComponent implements AfterViewInit, ControlValueAccessor {
...
-    this.isSelectedValue = true;
     this.isOpen = false;
     this.soundService.play(Sound.Check);
     this.onChangeCallback(selectedOption);
     this.optSelect.emit(selectedOption);
   }
 
-  toggle(e: any) {
-    e.stopPropagation();
+  getSelectedLabel(): any {
```
scomans, line 92:
```suggestion
  get selectedLabel(): string {
```

Code/skyrim_ui/src/app/components/dropdown/dropdown.component.ts
```diff
@@ -105,18 +112,21 @@ export class DropdownComponent implements AfterViewInit, ControlValueAccessor {
...
   }
 
   writeValue(obj: any): void {
     if (obj != null) {
-      this.isSelectedValue = true;
-      console.log(this.options.getValue(), obj);
       this.selected = this.options.getValue().findIndex(o => o.value === obj);
     } else {
-      this.isSelectedValue = false;
     }
```
scomans, line 130: remove else

Code/skyrim_ui/src/app/components/group/group.component.ts
```diff
@@ -125,24 +123,52 @@ export class GroupComponent implements OnInit, OnDestroy {
...
+  }
+
   ngOnDestroy() {
     if (this.timerSubscription) {
       this.timerSubscription.unsubscribe();
     }
   }
 
-  private flashGroup() {
+  private flashGroup() {
```
scomans, line 171:
```suggestion
  private flashGroup() {
```

scomans (changes requested):

Code/skyrim_ui/src/app/services/setting.service.ts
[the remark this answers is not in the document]
tomfull123: In the case you want to implement settings which don't use local storage you can just override the next function and not call the super definition.
Another solution would be to move the duplicated logic into a function on the base class which is specifically for saving to local storage?

scomans (approved):

tomfull123 (approved):
Looks good to me!

PR #543, opened 2022-12-12
feat(resources): cherry pick code.

Force67:
Implements the concept of resources for the server:

**Resource manifest spec**

The resource manifest is a ini file that describes the resource. It is used for keeping track of metadata about the resource, and for defining the (optional) scripting entry point. All resources haave to provide a manifest to avoid chaos and provide versioning.

**Resource structure**

Next to the server, there is the resources folder. The resource folder a folder for each resource and within it, its corresponding manifest.

```
└───resources
    ├──a
    │  └──resource-a.manifest
    └──b
       └──resource-b.manifest
```

**Manifest fields**

The order of the manifest entries does not matter. The manifest entries are case sensitive.

- `name`: The name of the resource.
- `version`: Your own version of the resource. Numeric and should follow [semver](https://semver.org/).
- `apiset`: Which version of the scripting api the resources subscribes to, in [semver](https:://semver.org).
- `entrypoint`: The name of the entrypoint script. This is optional, and defaults to nothing. The scripting system will determine the runtime based on the extension specified, so .lua files will be fed to a runtime that indicates it supports Lua, and so on.
- `dependencies`: A list of dependencies. The dependencies are specified as `resource_name@version`. The version is optional, and defaults to `*`. The version is specified in [semver](https://semver.org/). The dependencies are resolved in the order they are specified. If a dependency is not found, the resource will not be loaded. If a dependency is found, but the version is not compatible, the resource will not be loaded.
- `description`: A short description of the resource.
- `keywords`: (*Optional*) A list of keywords that describe the resource.
- `license`: (*Optional*) The license that the resource is released under.
- `repository`: (*Optional*) The URL of the repository that the resource is hosted in.
- `homepage`: (*Optional*) The URL of the resource's homepage.

**Example**

```
[Resource]
name = "my-resource"
version = 1.0.0
apiset = 1.0.0
description = "A resource that does stuff"
keywords = ["selfmade", "resource"]
license = "MIT"
repository = ""
homepage = ""
entrypoint = "test.lua"
dependencies = ["dependency-a@1.0.0", "dependency-b@1.0.0"]
```


Code/components/resources/SemanticVersion2.h
```diff
@@ -0,0 +1,128 @@
...
+    SemanticVersion(const underlying_type aVersionBits) noexcept
+        : m_versionBits(aVersionBits)
+    {
+    }
+    SemanticVersion(uint32_t aMajor, uint32_t aMinor, uint32_t aPatch) noexcept { From(aMajor, aMinor, aPatch); }
+    SemanticVersion(const char* apVersion) noexcept { FromString(apVersion); }
+
+    bool From(uint32_t major, uint32_t minor, uint32_t patch) noexcept
+    {
+        if (major > 4194302 || minor > 4194302 || patch > 4194302)
```
yamashi, line 31: Magic values would benefit from being constants

Code/components/resources/ResourceCollection.cpp
```diff
@@ -0,0 +1,292 @@
...
+void ResourceCollection::CollectResources()
+{
+    m_resourcePath = std::filesystem::current_path() / kResourceFolderName;
+    if (!std::filesystem::exists(m_resourcePath))
+    {
+        spdlog::info("Resource folder {} does not exist", m_resourcePath.string());
+        return;
+    }
+
+    std::vector<std::filesystem::path> manifestCanidates;
```
yamashi, line 239: TiltedPhoques::Vector

Code/components/resources/Manifest.h
```diff
@@ -0,0 +1,36 @@
...
+#include <filesystem>
+
+namespace Resources
+{
+namespace fs = std::filesystem;
+
+struct Manifest001
+{
+    /// name @ version
+    using DependencyTuple = std::pair<TiltedPhoques::String, SemanticVersion>;
```
yamashi, line 17: Make that a struct, pair is an antipattern

yamashi (changes requested):

Code/components/resources/ResourceCollection.h
```diff
@@ -9,7 +9,7 @@
 
 namespace Resources
 {
-class ResourceCollection
+struct ResourceCollection
 {
 public:
```
yamashi, line 14: useless public

Code/components/resources/SemanticVersion2.h
```diff
@@ -13,11 +13,14 @@ namespace Resources
 // 3 * 21 bits = 63 bits used for all three version numbers
 // the empty bit is used to double the range of each version number from 2097151 to 4194302
 // if it is set, every number starts at 2097151 instead of 0, but we only store the delta of (max) 2097151
-class SemanticVersion
+struct SemanticVersion
 {
 public:
```
yamashi, line 18: useless public

yamashi (changes requested):

yamashi (approved):

PR #552, opened 2023-01-03
Fix: equipment related crash

RobbeBryssinck:
ExtraDataList seems to be deallocated on a different thread sometimes. This seems to happen right between us fetching the inventory delta and locking the list. Not sure how Bethesda manages to avoid this, but calling the native UnequipAll function is safer than doing it manually anyway.

One user reported significantly fewer crashes when using this branch, possibly cutting down on a significant amount of crashing for other users as well.


Luc45:
howdy, my brother is crashing all the time. We constantly share equipments, because I'm the smither and he's the enchanter. Would we make sense that we try this build? What's the estimated release date for this to be included in a stable release?

Thanks.

PR #588, opened 2023-05-05
fix: SetTime crash

RobbeBryssinck:
ArgStack integer arguments cannot be popped with any other type than int64_t, as that's how the argstack is created, and std::any will throw if you try to cast it to any other type.

https://github.com/tiltedphoques/TiltedEvolution/blob/master/Code/components/console/ConsoleRegistry.cpp#L290

### DIFF
--- a/Code/server/GameServer.cpp
+++ b/Code/server/GameServer.cpp
@@ -327,12 +327,12 @@ void GameServer::BindServerCommands()
 
     m_commands.RegisterCommand<>("quit", "Stop the server", [&](Console::ArgStack&) { Kill(); });
 
-    m_commands.RegisterCommand<int, int>(
+    m_commands.RegisterCommand<int64_t, int64_t>(
         "SetTime", "Set ingame hour and minute", [&](Console::ArgStack& aStack) {
             auto out = spdlog::get("ConOut");
 
-            auto hour = aStack.Pop<int>();
-            auto minute = aStack.Pop<int>();
+            auto hour = aStack.Pop<int64_t>();
+            auto minute = aStack.Pop<int64_t>();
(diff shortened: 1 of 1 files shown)


Code/server/GameServer.cpp
```diff
@@ -327,12 +327,12 @@ void GameServer::BindServerCommands()
...
         "SetTime", "Set ingame hour and minute", [&](Console::ArgStack& aStack) {
             auto out = spdlog::get("ConOut");
 
-            auto hour = aStack.Pop<int>();
-            auto minute = aStack.Pop<int>();
+            auto hour = aStack.Pop<int64_t>();
+            auto minute = aStack.Pop<int64_t>();
             auto timescale = m_pWorld->GetCalendarService().GetTimeScale();
 
             bool time_set_successfully = m_pWorld->GetCalendarService().SetTime(hour, minute, timescale);
```
Zigler, line 338: Type down-conversion occurring here from int64_t to int which on most systems is int32_t.

Zigler: Furthermore , wouldn't an unsigned integer or std::ratio'd uint make more sense for a quantity of time since there isn't ever negative time in this instance?

RobbeBryssinck: @Zigler  Where is the type down conversion happening according to you? The integer on the stick is an int64_t, the popped value is an int64_t. Also, yes, semantically,  an unsigned integer would make more sense here, you could add another case for it in the builder, but it also doesn't really make a difference.

RobbeBryssinck: Oh I see, SetTime takes ints. Then yes, you are correct, a case should be added that matches the exact arguments of SetTime, which could indeed be changed to unsigned ints.

PR #600, opened 2023-06-19
Complete (Optional) Calendar Sync and Time Progression

KevinfromHP:
A while ago, Sturlen submitted a PR to sync up calendar dates (see #501). Their plan was to make the entire date sync up between players, but did not get to the networking part. The intentions of this PR are similar to theirs.

### Fixes
- Fixes the server forcing the date to be 0th day of the 0th year for all players.

### New features
- When a player joins the server, the server time syncs up to them. There is server config to sync the D/M/Y as well.
- When another player joins the server, the server checks to see whether the server or the newly connected player is further. It uses the one that is a further date for all connections.
- New command to set the date.

### Code Changes/Tweaks
- Moved TimeModel to Encoding, added serialization.
- Client now sends TimeModel as a part of its Authentication Request.
- ServerTimeSettings now uses TimeModel as a holder for Time info.
- Removed time resetting to offline time during disconnect. It causes a multitude of issues, primarily with daily spells and other things related to that. It's better to keep players at the time they are between loads rather than loading them back to their original state. It also caused problems when loading saves made while on the server.


KevinfromHP:
I should note the first commit's name "Time syncs to first player" is no longer accurate. It syncs to the player who has the furthest date, as mentioned.

Code/encoding/Structs/TimeModel.cpp
```diff
@@ -0,0 +1,106 @@
...
+void TimeModel::Update(uint64_t aDelta)
+{
+    float deltaSeconds = static_cast<float>(aDelta) / 1000.f;
+    Time += (deltaSeconds * (TimeScale * 0.00027777778f));
+
+    if (Time > 24.f)
+    {
+        int maxDays = GetNumerOfDaysByMonthIndex(Month);
+
+        while (Time > 24.f)
```
maximegmd, line 28: I am sure we can do better than this loop!

maximegmd (changes requested):

Code/encoding/Structs/TimeModel.cpp
KevinfromHP: Sure. I've got a few changes I need to push, but I didn't actually write that, it's been there. I'll look into it when I'm up tomorrow.

maximegmd (approved):

KevinfromHP:
Just need to add a command to set the date from the server and I think this'll be ready!

maximegmd:
> Just need to add a command to set the date from the server and I think this'll be ready!

Awesome! thanks for your work!

KevinfromHP:
Oh, that #include <format> is redundant. I'll get rid of that.

PR #603, opened 2023-06-23
Tweak: Allows Players to be invited to parties even if they are in a party.

KevinfromHP:
Good for when the party leader crashes and the party needs to be remade. Players no longer need to leave their current party to resync with the original party leader in this circumstance. The player leaves their old party when this happens.


Force67:
Do you intend on continuing this or can we close it?

PR #604, opened 2023-06-24
Upgrade UI dependencies

scomans:
This upgrades Angular to v16. No changes to the code were made.


RobbeBryssinck:
I don't really see the point in this. Does this fix anything that was previously broken?

scomans:
LTS of Angular 14 ends in november. Otherwise it is not really needed. Just wanted to keep it up to date as there are no breaking changes.

PR #605, opened 2023-07-01
Fix: Memory::Free id

RobbeBryssinck:
Apparently, the id we used for deallocations was the "Reallocate" id. We use the game's internal memory manager sparsely, so this could not have caused any major memory leaks, but a fix is still in order. The PR also includes a small script that matches idb names with codebase names for functions. This can be used to eyeball whether there are any other glaring mismatches. So far, I found one potential mismatch in quest code, which could be investigated.


Force67:
again, hardcoded paths

PR #606, opened 2023-07-01
Fix: quest start code

RobbeBryssinck:
Simplified the quest SetStage and start code. This should fix the bug where quests don't always get started immediately for party members when the leader gets it. Might also fix unrelated odd behavior and/or crashes, as a wrong function signature did not contain the correct parameters.


Force67:
Cool but the script contains hardcoded paths

PR #611, opened 2023-07-21
Combat targeting 2.0

RobbeBryssinck:
This PR does two things. Firstly, a new combat targeting system is implemented. This system (partly) reimplements the vanilla targeting system. It is cleaner than the initial solution, and should work better. Secondly, every second, for every remote Actor, the 3D is updated. Presumably, in the vanilla positioning code, Havok handles this. Since we use SetPosition directly, the "physics" entity does not always seem to update, causing a whole host of bugs, including (but not limited to) remote players not being targeted at all.


Code/client/Components/InterpolationComponent.h
```diff
@@ -21,6 +21,12 @@ struct InterpolationComponent
         TimePoint& operator=(const TimePoint&) = default;
     };
 
+    InterpolationComponent(std::chrono::steady_clock::time_point aPoint)
+        : last3DTimePoint(aPoint)
+    {}
+
+
+    std::chrono::steady_clock::time_point last3DTimePoint;
```
maximegmd, line 29: Code style

Code/client/Games/Skyrim/Combat/CombatController.cpp
```diff
@@ -0,0 +1,91 @@
...
+        ArrayQuickSortRecursiveCombatTargets(&targetSelectors, 0, targetSelectors.length);
+
+    pActiveTargetSelector = nullptr;
+    BSPointerHandle<Actor> newTarget{};
+
+    if (targetSelectors.length)
+    {
+        for (auto* pTargetSelector : targetSelectors)
+        {
+            if ((pTargetSelector->flags & 1) != 0 && (pTargetSelector->flags & 2) == 0)
```
maximegmd, line 35: if(pTargetSelector->flags & 3) == 1) ?

maximegmd (changes requested):

Code/client/Games/Skyrim/Combat/CombatController.cpp
RobbeBryssinck: The code in this function is a recreation of the original UpdateTarget function. The original code doesn't check for 3.

maximegmd (approved):

PR #640, opened 2024-02-22
Fix master build. Undo an apparently unused and inadvertent change in…

rfortier:
…troduced by commit 8d1130c2f77798ce15d892afbfa17d14001d64ac, that causes most of the client code to recompile on every compile.


RobbeBryssinck:
Already fixed in dev branch, will be committed later.

rfortier:
Thanks!

PR #641, opened 2024-02-29
If you ever suffer from recursive or looping crashes, this is the (trivial) patch for you

rfortier:
If there's a crash cycle or recursive cycle, "the real problem crash dump" is overwritten by a "subsequent problem crash dump," hiding the real problem. And the loop may go on forever. Stop after first crash dump.


rfortier:
Took a look at the failing build log. It's environmental outside if the single-file the PR changes.

PR #642, opened 2024-03-13
Fix game exit crash when under ModOrganizer2, only Address Library

rfortier:
… and STR mods are loaded

Under MO2, there's a bug caused when calling RealGetModuleFileNameW on XAudio2_7.dll, it crashes in usvfs. This only happens during the game shutdown paths when the last few threads are exiting. This PR just avoids that bug.

TODO: Further analysis of what is under MO2 USVFS and what is needed.


rfortier:
Some additional color commentary. This isn't the greatest fix, but XAudio2_7.dll (a vanilla DirectX component) provokes a usvfs crash at exit time when running under ModOrganizer2. Even with no mods, the only things loaded are Address Library and STR. It isn't even necessary to load a game, just run the launcher and select quit.

The DLL is opaque. We just need a couple more threads to exit. So chalk this up to another Bethesda Band-Aid.

PR #643, opened 2024-03-13
Fix game exit crash when under ModOrganizer2, when only Address Library

rfortier:
…and STR mods are loaded

Under MO2, there's a bug caused when calling RealGetModuleFileNameW on XAudio2_7.dll, it crashes in usvfs. This only happens during the game shutdown paths when the last few threads are exiting. This PR just avoids that bug.

TODO: Further analysis of what is under MO2 USVFS and what is needed.

### DIFF
--- a/Code/immersive_launcher/stubs/FileMapping.cpp
+++ b/Code/immersive_launcher/stubs/FileMapping.cpp
@@ -218,7 +218,15 @@ DWORD WINAPI TP_GetModuleFileNameA(HMODULE aModule, char* alpFileName, DWORD aBu
     ScopedOSHeapItem wideBuffer((aBufferSize * sizeof(wchar_t)) + 1);
 
     wchar_t* pBuffer = static_cast<wchar_t*>(wideBuffer.m_pBlock);
-    DWORD result = RealGetModuleFileNameW(aModule, pBuffer, aBufferSize * sizeof(wchar_t));
+
+    // Under MO2, there's a bug caused when calling RealGetModuleFileNameW on XAudio2_7.dll,
+    // it crashes in usvfs. This only happens during the thread shutdown paths 
+    // of quitting, so just avoid the bug. 
+    // TODO: Further analysis of what is under MO2 USVFS and what is needed.
+    DWORD result = 0;
+    if (aModule != GetModuleHandleW(L"XAudio2_7.dll"))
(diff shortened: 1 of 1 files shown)


rfortier:
Some additional color commentary. This isn't the greatest fix, but XAudio2_7.dll (a vanilla DirectX component) provokes a usvfs crash at exit time when running under ModOrganizer2. Even with no mods, the only things loaded are Address Library and STR. It isn't even necessary to load a game, just run the launcher and select quit.

The DLL is opaque. We just need a couple more threads to exit. So chalk this up to another Bethesda Band-Aid.

Dragonisser:
Why are you keep opening pull request

rfortier:
> Why are you keep opening pull request

Because I didn't do it right, and cosi told me to put it on the /dev branch

Dragonisser:
Then just change the branch and dont create new pull request

rfortier:
Cosi observed something to the effect _that looks like a band-aid, can't we find the real cause and fix it?_

So I dug back on my investigations, and found the link describing the bug. It is a bug in XAudio2_7.dll. The bug is not in XAudio2_8.dll, but isn't compatible, and XAudio2_9.dll introduced a compatibility mode that has to be requested. This was the shorter fix.

I'll link the blog with gory details, but links sometimes die so here's the synopsis: the crash occurs if you try to unload XAudio2_7.dll, typically during the exit path (like we're, or MO2 USVFS, is doing) while there are still any active objects in the library. The workaround he proposes is to create an extra reference to the DLL so it doesn't unload; you can delete your extra reference when it is "known safe" to do so.

The PR approach is simpler and very localized; if trying to unload XAudio2_7.dll (specifically), just skip it. We're in the exit path anyway, let exit clean it up.

We could try and do what the blog suggests, but it is not clear where a "known safe" spot would be. Why can't audio still be playing when you quit? It would take a fair amount of research and introduction of riskier code to do so. 

Another more complex workaround might be to exclude XAudio2_7.dll, or all Windows DLLs, from USVFS processing. Again, more complicated and riskier in the final stages of release train. But if someone knows how to do that, let's talk and there can be a follow-on.

Here's the Microsoft link (but note it redirects to an individual's blog, that's expected): [Fix for XAudio2_7.dll crash](https://blogs.msdn.microsoft.com/chuckw/2015/10/09/known-issues-xaudio-2-7). The link is bugged and doesn't go straight to the section header, so search for _Xaudio2_

PR #644, opened 2024-03-13
If you ever suffer from recursive or looping crashes, this is the (tr…

rfortier:
…ivial) patch for you.

If there's a crash cycle or recursive crash, "the real problem crash dump" is overwritten by a "subsequent problem crash dump," hiding the real problem. And the loop may go on forever. Stop after first crash dump.

### DIFF
--- a/Code/client/CrashHandler.cpp
+++ b/Code/client/CrashHandler.cpp
@@ -22,7 +22,9 @@ std::string SerializeTimePoint(const time_point& time, const std::string& format
 
 LONG WINAPI VectoredExceptionHandler(PEXCEPTION_POINTERS pExceptionInfo)
 {
-    if (pExceptionInfo->ExceptionRecord->ExceptionCode == 0xC0000005)
+    static int alreadycrashed = 0;
+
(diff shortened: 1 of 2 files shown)


Dragonisser:
Wrong branch

rfortier:
OK, moved to /dev branch. Next time I'll know.

rfortier:
This PR doesn't change anything in imgui, so that compile eror is elsewhere.

RobbeBryssinck:
The latest dev branch needs to be merged with this PR first, as the GLM issue has been fixed there, and the additional define here is not required anymore.

As for the fix itself: there was a bug in the latest distribution where IS_MASTER was not defined. This meant that debug menus, the console and the local crash handler referenced here were included. The crash handler won't be included in 1.6. Still for debug purposes, this PR can be pushed through.

rfortier:
With the update to xmake /dev branch, this should compile and be good to merge now. 

I just rebuilt a full releasedbg build, which still includes the code, on today's dev branch and it compiles fine.

rfortier:
Updated the PR to revert the added #define as it is no longer needed.

The build is failing on the gamenetworkingsockets / protobuf / xmake fix we're all waiting on.

PR #658, opened 2024-04-16
Remove limit to BehaviorVars on the wire.

rfortier:
STR is up against the limit, using 62 of a hardwired 64  boolean behavior vars to be synced. This fix removes the limit before it becomes a barrier.

Since "unlimited" will bite us, note the base code has some lengths encoded in 16-bit ints that will bite eventually. Trying to sync even a fraction of that many will cause problems, though.

Reimplemented the Boolean long long as a (bit) Vector. This made all vars (bools, ints, floats) vectors, so some code simplification of the wire protocol is accessible.

The Booleans are just passed "as is," no attempt to detect which ones changed as the space required to send the delta is more than just sending the bits.

Overall impact on wire protocol: sending humanoids will be slightly longer, as the 64-bits available is almost full. But that's longer by a byte (or maybe two since I used String to simplify the code).

Everything else sent will be shorter and less overhead.

There's no avoiding iterating the bits of the Vector<bool>. But the overhead when translating the array of \<bool> to an array of bits and back is about the same.

There was a bunch of extra overhead if Serialization::WriteBool() were to be used, so optimized that part out.

Overall, should be better on the wire, and neutral-to-better on CPU.

2nd commit msg:
Reworked creating and applying Behavior variable diffs and serializing them. Simpler, easier to read, should usually save bytes on the wire.

### DIFF
--- a/Code/client/Games/References.cpp
+++ b/Code/client/Games/References.cpp
@@ -227,10 +227,9 @@ void TESObjectREFR::SaveAnimationVariables(AnimationVariables& aVariables) const
             if (!pVariableSet)
                 return;
 
-            aVariables.Booleans = 0;
-
-            aVariables.Floats.resize(pDescriptor->FloatLookupTable.size());
-            aVariables.Integers.resize(pDescriptor->IntegerLookupTable.size());
+            aVariables.Booleans.assign(pDescriptor->BooleanLookUpTable.size(), false);
+            aVariables.Floats.assign(pDescriptor->FloatLookupTable.size(), 0.f);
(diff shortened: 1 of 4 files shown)


rfortier:
This PR changes the wire protocol, so it is worth including in the release. It also has a fair amount of play testing on it and has proven to be solid, so I think it is reasonably safe to include. 

It is technically not a prerequisite for the modded animation support to be PR'ed separately. But that PR wouldn't be much use without it.

rfortier:
This PR was combined into PR #659

PR #659, opened 2024-04-25
Feature: support for behavior mods

rfortier:
Behavior mods such as those generated by Nemesis or Pandora.

Important note: this PR supports conditional compilation, _which is currently set to off_ in the last commit. This was done to enable the code to merge without risking the imminent release. If the conditional compile is switched on, the support will be enabled in the clients and those clients should be able to use the same servers as clients without the support. 

**Goals**
- Not risk the release.
- Enable compilation of a "pre-release" or test version that could be compatible with the public game servers.

**Prerequisites**
- Previously, PR #658 was a prerequisite. Cosi asked me to merge that branch, so this is no longer true.

**Testing**

Four combinations were tested to ensure the PR is safe:
1. Compiled off, no mods: animations sync as expected.
2. Compiled off, with behavior mods: behaviors not synced, characters skating, as expected.
3. Compiled on, no mods: animations sync as expected.
4. Compiled on, with behavior mods: behaviors sync for the mods supported in the PR. Support for additional simple mods that just need an added variable or two synced can be added by dropping the definitions into the SkyrimTogetherRebornBehaviors directory.

### Theory of Operation (how it works)

It's a big PR in line count, but much of that is machine-translated code. The major topics to cover are:

1. Directory layout, source directories and installation directories.
2. How the mod accomplishes its goals of enabling simple animations (no complex event management or scripting) generated by Nemesis or Pandora to sync.
3. Approaches to enable STR devs and modders, even end users to make behavior changes without stepping on each other.
4. Support for syncing additional behavior variables (in particular, more than 64 boolean variables, of which the STR code is already using 62). Note this requires a protocol change between the server and clients, so this code _always_ compiles, while the remainder is conditionally compiled. This approach simultaneously supports two goals:  the STR team can release without the mod support for lowest risk, but a version not supported by the STR team can also be compiled, and use the same servers.

### Directory Layout / Changes

**Source Directories**
- One source directory is added to the source tree at _client\ModCompat_. The changes support both Skyrim and Fallout4.
- One file, an encumbrance, is added at _\Libraries\magic_enum.hpp_. This enables translating enum names to their string form, despite C++ not supporting reflection (yet). It is governed by the [MIT license](http://opensource.org/licenses/MIT).
- A directory tree is added to _\Gamefiles\Skyrim\SkyrimTogetherRebornBehaviors_. The files in this directory enable the STR mod to sync additional behavior/animation variables if a mod requires it. Any variables named in the tree are merged with the base variables the STR dev team defines. Additional files define the original STR hash of behavior variables, and a _signature_, a unique list of behavior variables that must exist (or must not exist) to uniquely identify an Actor.

**Installation Directories**

- The source directory _\Gamefiles\Skyrim\SkyrimTogetherRebornBehaviors_ must be installed alongside the _SkyrimTogetherReborn_ directory by any FOMOD or 7z installer. It is deliberately outside the SkyrimTogetherReborn directory so 3rd party tinkering is outside of the game directory.

### Theory of Operations

In a sentence, the mod works by detecting _AnimationGraphDescriptor_ hashes that don't exist in base STR, seeing if the Actor's animation variables match a signature in the _SkyrimTogetherRebornBehaviors_ directory, and if so, merge the STR variables with the ...Behaviors directory variables (if any, it might just be the hash was changed), and generates a new AnimationGraphDescriptor with the new hash. 

The interception requires a few lines of code in _References.cpp_ to intercept where animation variables are prepared for serialization and deserialization. They call out to _BehaviorVar.cpp:BehaviorVar::Patch()_ when the AnimationGraphDescriptor lookup-by-hash fails.

::Patch can fail and return NULL again. This is normal; it happens when STR doesn't need to sync behaviors, syncing position and velocity is enough, so there is no AnimationGraphDescriptor at all for that creature. Wisps are an example.

If a hash lookup fails, it is added to a failed list and ignored for 10m. This because searching for a signature is semi-expensive, but more because it generates a ton of logs that aren't useful which is the real cost. 10m is long enough for most encounters to end, but short enough to try again if dynamic behavior modification is ever properly supported.

**Intricacies of Merging Behavior Var Lists**

Nemesis doesn't just add more behaviors to the end of the list; it _rearranges_ the list. That's unfortunate, and it is why humans skate if just Nemesis is installed without adding any more variables to be synced; The rearranged list has a different hash. 

To accomplish a merge, these operations occur:

1. We must be able to translate the base STR _AnimationGraphDescriptors_ back to their string form. This is done at authoring time by running a script that translates the AnimationGraphDescriptor* files to a form that _magic_enum_ can use to generate the string form of each enum, for everything in the Skyrim and Fallout4 directories. Currently, the output files are checked into Git. It would be better to get the _xmake_ dependencies right, but I'll need some advice for that. And it can wait until after release.
2. At game startup time, the _SkyrimTogetherRebornBehaviors_ directory is scanned. Each subdir is expected to have a _*hash.txt_ file with the original STR base game hash, and a _*sig.txt_ file that contains a list of behavior variables that must and must not exist to uniquely identify the Actor. Additional _*bool.txt_, _*int.txt_, and _*float.txt_ files name any additional variables that must be synced. Each mod author can have their own files with a unique prefix, they are all merged and stored in a list of BehaviorVar::Replacer objects.
3. At run time, the modded behavior is matched to a _Replacer_ via the unique signature.
4. The _Replacer_ identifies the matching STR base game _AnimationGraphDescriptor_, which is translated back to strings which are stored in a ::Set for each of bool, int and floats.
5. Added variables (if any) found in the _BehaviorVar::Replacer_ built from the _SkyrimTogetherRebornBehaviors_ tree are merged (uniquely, because of the way ::Sets work).
6. The sets are translated back to an enum form to construct a new AnimationGraphDescriptor.
7. That descriptor is added to the game data structures in _AnimationGraphDescriptorManager::Register_

From that point forward, the new hash will be successfully found by the STR base-game logic.

There are a few details that added some complexity. In particular, there are misspellings of variable names even in vanilla  Skyrim, usually incorrect case like _speed_ when the correct variable is _Speed_; that happens with a particular Winterhold guard, for example. Having no way to know if Skyrim ever uses any variables only distinguished by capitalization, the code tries a few common variants if a lookup fails and generally will correct the problem. 

**Decoupling STR dev team and Modders**

All this logic to match and merge is to ensure the STR dev team and modders can make changes independently without interfering with each other.

**Support for Syncing Additional Behavior Variables**

All of the above support will ensure animation sync if Nemesis/Pandora are used, and if no new variables need to be synced will even be compatible with the same servers used by clients without this animation support. 

But to also support mods that do need to sync additional variables, such as the popular [True Directional Movement](https://www.nexusmods.com/skyrimspecialedition/mods/51614), this mod needs support for syncing more variables. In particular more Boolean variables, STR base uses 62 out of a limit of 64. The limit is baked into protocol between the clients and server, so lifting the limit creates a compatibility problem.

This PR includes a merge of branch _feat-unlimited-behavior-vars_, which always compiles once this PR is merged to ensure versions compiled with behavior mod support will still be able to use the public servers of the same version.

Rather than simply turning the boolean long long in the code into an array, this code changes the definition _AnimationVariables::Booleans_ from a uint64_t to a _TiltedPhoques::Vector< bool >_. This compiles into an array of bits with the current tech stack (this isn't _guaranteed_ by the C++ standard, but is the norm). 

This design choice was made because:
- It actually takes 1 byte less on the wire than the than the approach of just adding another uint64_t in the "large" case of sending humanoids.
- It takes less space on the wire than the former protocol for most creatures that don't sync very many variables. 
- Consistent code, everything is a Vector now.


rfortier:
Some more info for people interested in testing this, but not wanting to bother with compilation.

Please report issues [here ](https://github.com/rfortier/TiltedEvolution-rwf/issues) on github, not to the STR dev team. Or find Ujave on the STR Discord.

I have compiled up two versions of this mod and released it on my github; a version with the compilation flag turned on, and a version with it compiled off for testing that it is safe. There's a FOMOD installer provided so anyone can try them. You can find that [here](https://github.com/rfortier/TiltedEvolution-rwf/releases/tag/v0.22-FOMOD-STR1.6.Xasof20240331-Feat-Integration-86cfa602).

There's not a lot of documentation yet about how to support additional mods, but @MostExcellent or I can help.

rfortier:
As requested, I merged the content of PR #658 into this PR. 

Support for behavior mods is _compiled off_ by default.
The former content of PR #658 always compiles.

I'll close PR #658 now.

Code/client/ModCompat/magic_enum.hpp
RobbeBryssinck, line 1: This should be a library (in the Libraries folder).

rfortier: >**RobbeBryssinck**: What's missing from this PR is a general overview of the PR's contents and how you've solved the problem at hand. Currently, I'd have to read the code to do that, which is a bit much given the size of the PR.

I'll do that today. A Theory of Operation type overview, right? It will be a couple of pages or so, in an edited PR description.

>**RobbeBryssinck**: This [magic_enum.hpp] should be a library (in the Libraries folder).

I'll move it. The original thought was to keep it self-contained, but if it's going to be merged I understand it should be where it belongs.

RobbeBryssinck: Yeah just a conceptual overview of how the new system works and how it compares to the old system would be great.

rfortier:
Rebased to /master for release.

rfortier:
Updated branch to the latest /dev.

It currently compiles with MODDED_BEHAVIOR_COMPATIBILITY undefined. So only Unlimited Vars is compile, making a standard build server-compatible with a modded animation build.

Revert the last commit to build an animation build.

Code/client/ModCompat/BehaviorVarsMap.h
Force67, line 1: Spacing, also we do not use std::map, use the tiltedcore map, as std::map is extremely slow.

Code/client/ModCompat/BehaviorVarsMap.cpp
```diff
@@ -0,0 +1,51 @@
+#ifdef MODDED_BEHAVIOR_COMPATIBILITY
+#include "BehaviorVarsMap.h"
+
+
+uint32_t BehaviorVarsMap::find(const uint64_t acKey, const std::string acName)
+{
+    auto map = m_map.find(acKey);
```
Force67, line 7: a better variable name would be nice + spacing everywhere is off (empty lines)

Force67: We have a clang format you can execute over the codebase

Code/client/ModCompat/BehaviorVar.h
```diff
@@ -0,0 +1,59 @@
...
+    const AnimationGraphDescriptor* constructModdedDescriptor(
+        const uint64_t acNewHash, const Replacer& acReplacer,
+        std::map<const std::string, const uint32_t>& acReverseMap);
+
+    std::vector<std::filesystem::path> loadDirs(const std::filesystem::path& acPATH);
+    Replacer* loadReplacerFromDir(const std::filesystem::path acDir);
+    std::vector<uint64_t> signatureMatches(const uint64_t acHash, const std::string acSignature) const;
+
+    std::vector<Replacer> behaviorPool; // Pool for loaded behaviours
+    std::map<uint64_t, std::chrono::steady_clock::time_point> failedBehaviors;
```
Force67, line 53: Tiltedcore map

Code/client/ModCompat/BehaviorVar.h
```diff
@@ -0,0 +1,59 @@
...
+    const std::vector<std::string> tokenizeBehaviorSig(const std::string acSignature) const;
+    const AnimationGraphDescriptor* constructModdedDescriptor(
+        const uint64_t acNewHash, const Replacer& acReplacer,
+        std::map<const std::string, const uint32_t>& acReverseMap);
+
+    std::vector<std::filesystem::path> loadDirs(const std::filesystem::path& acPATH);
+    Replacer* loadReplacerFromDir(const std::filesystem::path acDir);
+    std::vector<uint64_t> signatureMatches(const uint64_t acHash, const std::string acSignature) const;
+
+    std::vector<Replacer> behaviorPool; // Pool for loaded behaviours
```
Force67, line 52: Tiltedcore vector

Code/client/ModCompat/BehaviorVar.h
```diff
@@ -0,0 +1,59 @@
...
+        std::set<uint32_t>& aBoolVars, 
+        std::set<uint32_t>& aFloatVars,
+        std::set<uint32_t>& aIntVars);
+
+    const std::vector<std::string> tokenizeBehaviorSig(const std::string acSignature) const;
+    const AnimationGraphDescriptor* constructModdedDescriptor(
+        const uint64_t acNewHash, const Replacer& acReplacer,
+        std::map<const std::string, const uint32_t>& acReverseMap);
+
+    std::vector<std::filesystem::path> loadDirs(const std::filesystem::path& acPATH);
```
Force67, line 48: Coding conventions? no camelCase

Code/client/ModCompat/BehaviorVar.cpp
```diff
@@ -0,0 +1,754 @@
...
+#include <mutex>
+std::mutex mutex_lock;
+
+// How long we failList a behavior hash that can't be translated.
+// See explanation in BehaviorVar::Patch
+using namespace std::literals;
+const std::chrono::steady_clock::duration FAILLIST_DURATION(10min);
+
+BehaviorVar* BehaviorVar::single = nullptr;
+BehaviorVar* BehaviorVar::Get()
```
Force67, line 52: Cant we store the instance with entt?

Code/client/ModCompat/BehaviorVar.cpp
```diff
@@ -0,0 +1,754 @@
...
+#include <Camera/PlayerCamera.h>
+#endif
+
+#include <mutex>
+std::mutex mutex_lock;
+
+// How long we failList a behavior hash that can't be translated.
+// See explanation in BehaviorVar::Patch
+using namespace std::literals;
+const std::chrono::steady_clock::duration FAILLIST_DURATION(10min);
```
Force67, line 49: Could this be constexpr?

Code/client/ModCompat/BehaviorVar.cpp
```diff
@@ -0,0 +1,754 @@
...
+#include "BehaviorVarsMap.h"
+
+#if TP_SKYRIM64
+#include <Structs/Skyrim/AnimationGraphDescriptor_Master_Behavior.h>
+#include <Camera/TESCamera.h>       // Camera 1st person is only in Skyrim?
+#include <Camera/PlayerCamera.h>
+#endif
+
+#include <mutex>
+std::mutex mutex_lock;
```
Force67, line 44: No global pls

Code/client/ModCompat/BehaviorVar.cpp
```diff
@@ -0,0 +1,754 @@
...
+{
+    std::string lowerCaseStr(acStr);
+    std::transform(acStr.begin(), acStr.end(), lowerCaseStr.begin(),
+                   [](unsigned char c) { return std::tolower(c); });
+    return lowerCaseStr;
+}
+
+// Converts the keys of a map (which are const std::string) to lowercase.
+// This is our "Plan C" if a vanilla var isn't found to make sure it isn't just a case-sensitivity issue.
+void lowerCaseKeys(const std::map<const std::string, const uint32_t>& acMap,
```
Force67, line 81: Tiltedcore types

Code/client/ModCompat/BehaviorVar.cpp
```diff
@@ -0,0 +1,754 @@
...
+        commaPos = acSignature.find(',', offset);
+        retVal.push_back(acSignature.substr(offset, commaPos));
+        offset = commaPos + 1;
+
+    } while (commaPos != std::string::npos);
+
+    return retVal;
+}
+
+const AnimationGraphDescriptor* BehaviorVar::constructModdedDescriptor(const uint64_t acNewHash, const Replacer& acReplacer, std::map<const std::string, const uint32_t>& acReverseMap)
```
Force67, line 216: Can we make the debug logging conditional somehow?

Code/client/ModCompat/BehaviorVar.cpp
```diff
@@ -0,0 +1,754 @@
...
+    // (possibly negated) signaturevar, followed by zero or more comma separated additional
+    // signaturevars (possibly negated).
+    //
+    // That's enough info to create a unique signature in the base game descriptors. At least,
+    // it was at the time of writing.
+    // 
+    // O(N) loop, but this typically only gets executed a few times (once for each modded
+    // creature TYPE, not each modded creature).
+    //
+    std::vector<size_t> matchedReplacers;
```
Force67, line 388: tiltedcore types

Code/client/Games/References.cpp
```diff
@@ -219,6 +224,13 @@ void TESObjectREFR::SaveAnimationVariables(AnimationVariables& aVariables) const
 
             auto pDescriptor = AnimationGraphDescriptorManager::Get().GetDescriptor(pExtendedActor->GraphDescriptorHash);
 
+#ifdef MODDED_BEHAVIOR_COMPATIBILITY
+            // Modded behavior check if descriptor wasn't found
+            extern const AnimationGraphDescriptor* BehaviorVarPatch(BSAnimationGraphManager * pManager, Actor * pActor);
```
Force67, line 229: Maybe im an idiot but wasnt that extern fn already declared at top of the file?

Code/client/ModCompat/BehaviorVar.cpp
```diff
@@ -0,0 +1,754 @@
...
+// This is our "Plan C" if a vanilla var isn't found to make sure it isn't just a case-sensitivity issue.
+void lowerCaseKeys(const std::map<const std::string, const uint32_t>& acMap,
+                   std::map<const std::string, const uint32_t>& aLowerCaseMap)
+{
+    for (const auto& item : acMap)
+        aLowerCaseMap.insert({toLowerCase(item.first), item.second});
+}
+
+// Process a set of variables, adding them to the aVariableSet
+void processVariableSet(const std::map<const std::string, const uint32_t>& acReverseMap,
```
Force67, line 89: Needs tiltedcore types

Code/client/ModCompat/BehaviorVar.cpp
```diff
@@ -0,0 +1,754 @@
...
+
+    // Prepare reading files
+    std::string sigVar;
+    std::string creatureName = aDir.filename().string();
+    std::vector<std::string> floatVar;
+    std::vector<std::string> intVar;
+    std::vector<std::string> boolVar;
+    std::string tempString;  // Temp string
+
+    std::ifstream fileSig(signatureFile);
```
Force67, line 541: You cannot assume that this succceeds, handle error.

Code/client/ModCompat/BehaviorVar.cpp
```diff
@@ -0,0 +1,754 @@
...
+    result->syncFloatVar   = floatVar;
+    result->syncIntegerVar = intVar;
+
+    return result;
+}
+
+// Find any behaviors which match the signature.
+// There should be exactly one.
+//
+std::vector<uint64_t> BehaviorVar::signatureMatches(const uint64_t acHash, const std::string acSignature) const
```
Force67, line 631: String by copy is a bad idea. maybe a string view?

Code/client/ModCompat/BehaviorVar.h
```diff
@@ -0,0 +1,59 @@
...
+struct BehaviorVar
+{
+    struct Replacer
+    {
+        uint64_t origHash;
+        uint64_t newHash;
+        std::string signatureVar;
+        std::string creatureName;
+        std::vector<std::string> syncBooleanVar;
+        std::vector<std::string> syncFloatVar;
```
Force67, line 15: No std types. use tiltedcore.

Code/client/ModCompat/BehaviorVar.h
```diff
@@ -0,0 +1,59 @@
...
+        std::vector<std::string> syncIntegerVar;
+        bool operator==(const uint64_t& acRHS) const
+        {
+            return origHash == acRHS;
+        }
+    };
+
+    static BehaviorVar* Get();
+    const AnimationGraphDescriptor* Patch(BSAnimationGraphManager* apManager, Actor* apActor);
+    boolean failListed(const uint64_t acHash);
```
Force67, line 25: bool?

Force67 (changes requested):

rfortier:
I'll run through your review comments. Tried to follow the style guide, but at least missed capitalized funcs/methods. Time to figure out  the clang formatter, I guess. 

Looks like a VS reinstall lost my tabs->4spaces setting. Is that the source of the generic "spacing" comment? And will clang formatter fix it? Otherwise I can do it manually.

Looks like I missed converting some contributed code to TiltedPhoques::. But one is deliberate; if I remember right TiltedPhoques::Map is a hashmap, and I needed a sorted map; is there one? NM, just found it.

Performance isn't super-critical for the maps, as they are only used once per creature type per launch (as in, once for all humanoids, once for all dragons...). So even if TP::SortedMap is slower, it won't hurt anything.

Code/client/ModCompat/BehaviorVarsMap.h
rfortier: Commit 2d989d98047dfa179045b7ba45333d5f17575a1b addresses most comments, other than clang-format will be an isolated commit to follow. Pushing this now to check CI (formatting _shouldn't_ break CI) and check a few things.

Common Fixes:
- Fixed all gaps in naming conventions; biggest one missing that FunctionsWereCapitalized(). 
- Replaced all std:: with TiltedPhoques:: functions _when possible._ Sometimes TP didn't have the function (sorted sets), and sometimes conversions were broken with no workaround.
- Final commit will enable the conditionally compiled code, as that seems to be the consensus.
- Note enabling the code introduces a README-ANIMATION-MODS that will need review. 
- There were a few call-by-values that were fixed to call-by-reference to avoid copies.

Code/client/ModCompat/BehaviorVarsMap.cpp
rfortier: Commit 2d989d98047dfa179045b7ba45333d5f17575a1b

Code/client/ModCompat/BehaviorVar.h
rfortier: Commit 2d989d98047dfa179045b7ba45333d5f17575a1b

Code/client/ModCompat/BehaviorVar.h
rfortier: Commit 2d989d98047dfa179045b7ba45333d5f17575a1b

Fixed it. It's a pretty narrow window, though. The path is from a directory iterator just run; so either the file was just deleted in the last millisecond or the directory is read-protected; a lot more would break in the latter case.

Code/client/ModCompat/BehaviorVar.h
rfortier: Commit 2d989d98047dfa179045b7ba45333d5f17575a1b

Code/client/ModCompat/BehaviorVar.cpp
rfortier: Commit 2d989d98047dfa179045b7ba45333d5f17575a1b

Code/client/ModCompat/BehaviorVar.cpp
rfortier: Commit 2d989d98047dfa179045b7ba45333d5f17575a1b

Fixed it. It's a pretty narrow window, though. The path is from a directory iterator just run; so either the file was just deleted in the last millisecond or the directory is read-protected; a lot more would break in the latter case.

Code/client/ModCompat/BehaviorVar.h
rfortier: Commit 2d989d98047dfa179045b7ba45333d5f17575a1b

Code/client/ModCompat/BehaviorVar.h
rfortier: Pushed the clang-format commit. Reverted the "disable MODDED_ANIMATION_COMMIT".

So I think this is ready for re-review and merge.

Issue #656, opened 2024-04-14
Make the launcher file selection only allow "SkyrimSE.exe"


Dragonisser:
## Description

Since people cant follow the clearly written guide and cant read the window name saying "Select SkyrimSE.exe", it would be handy to just limit it to the exact file name. Both Steam and GOG have the same file name.

## Additional context

Hogwarp uses this limitation already.

https://github.com/tiltedphoques/TiltedEvolution/blob/master/Code/immersive_launcher/oobe/PathSelection.cpp#L41

PR #661, opened 2024-05-03
tweak: add TARGET_NAME to selectable-executable filter

ToeKneeRED:
Adds TARGET_NAME to the executable elements to select from in the executable-select dialog 
Related feature request https://github.com/tiltedphoques/TiltedEvolution/issues/656


RobbeBryssinck:
Maybe it'd be an idea to automatically select the executable, and only if it doesn't find it, show the dialogue?

RobbeBryssinck:
Sorry to yank your chain, but on second thought, having the dialogue pop up always is probably a better idea, since people with modded setups often have multiple installs. You can revert the last commit.

Issue #404, opened 2022-07-25
Automatic party invites 


meshik:
## Description

Crashes are inevitable. A tedious aspect of this involves getting an invite again from the party leader, who may be busy in dangerous combat from the time before you crashed. If I could get a party invite automatically as I connect, I can seamlessly re-join the battle.
This feature is ideal for private servers of consistent friends (of which there are many).

## Implementation

Launching a party in an auto-invite-enabled server makes it so any subsequent connection receives an invite to this party. Add a toggleable feature to the server (default is off) where one can specify if auto-invites should be used.

RobbeBryssinck:
Tbh if we're gonna go this route, it'd be better to just auto join a party on joining a server, instead of auto inviting someone. Takes away another step of having to go through the menus to accept an invite.

KevinfromHP:
I'll look into creating an auto-party config. I was already in the process of something similar but this sounds easier

Dragonisser:
Added in 1.6.5

Issue #378, opened 2022-07-22
Preset Parties on server


BobFigGreen:
Add option to create preset parties in server options
Due to crashes at the moment inviting people to parties all the time can become tedious
Would be very grateful if this feature could be implemented

BobFigGreen:
Or possibly add option to set a party leader for whole server

Sturlen:
Would be difficult to do, since you don't have a permanent user id or a login for the server. Right now the only thing the server has is your player characters name, and anyone can log in with any character name or even the same one.

BobFigGreen:
> 
True but even adding an option using character names would be fine for private servers as it would only need changing when making new characters 

I understand that this may not be possible though

Gibbit2k:
This should be achievable and really useful on small private servers:

1. Add a party leader line to the server config where the party leader nickname is specified
2. When connecting to server beside ip ask for a nickname 
3. If nickname is the one specified in server config connect to server and create party
4. If nickname is not the one in server config check if party leader is on, if true connect and join party automatically if not don't allow connect to server until party leader is on
5. Alternatively you could force dc all users if the party leader crashes or disconnects or something to try and mitigate sync errors.

Owlchemist:
> Would be difficult to do, since you don't have a permanent user id or a login for the server. Right now the only thing the server has is your player characters name, and anyone can log in with any character name or even the same one.

Well, if someone wanted to troll you, there are _far worse_ things they can do than grief your party composition. As I understand it the main development direction was playing with your friends, I think it'd be a nice bit of polish for non-public servers at least.

Sturlen:
It would still generate a lot of edge cases that need to be handled, even
if noone is aiming to be malicious.

I think it would be better to focus on making making inviting people
easier, such as clicking on their name in chat and being able to accept
without going into menus.

jjnether:
Best solution would be the feature mentioned in #404

isaaclw:
Agreed. If there are no parties, then the auto-join or auto-invite system can be turned off. The majority of the bugs happen with non-party members anyway (ie: crafting/inventory)

Dragonisser:
Added in 1.6.5

PR #663, opened 2024-05-04
Feature: auto party join

Chaybugz:
If there are no parties on the server, when a party is created, every player is automatically added to the party

If there is one party in the server, when a player connects, they are automatically added to the party

Can be toggled in the server config (on by default)

For issues #378 and #404


BobFigGreen:
Appreciate this

PR #665, opened 2024-05-05
Default enabled mod policy with vanilla mod list

RobbeBryssinck:
Now, by default, the server will check and enforce whether you really _really_ disabled all mods and CC content as instructed per the wiki ;)


VENDETTA5:
Imagine the amount of servers I won't be able to join because it will be enabled by default, lol.

But you know, if you like it, do it. I understand what you're coming from. I just hope will not be too lazy to change that setting when needed.

PR #682, opened 2024-05-25
fix quest start

otsffs:
Addressing synchronization issues during the initialization phase of partial sub-tasks.


Code/server/Services/QuestService.cpp
```diff
@@ -44,6 +44,8 @@ void QuestService::OnQuestChanges(const PacketEvent<RequestQuestUpdate>& acMessa
                 spdlog::debug("Started quest: {:X}:{:X}, stage: {}", message.Id.ModId, message.Id.BaseId, message.Stage);
 
                 notify.Status = NotifyQuestUpdate::Started;
+            }else{
```
RobbeBryssinck, line 47: Please keep styling in line with the rest of the code (else and curly brackets on separate lines).

RobbeBryssinck (changes requested):
Good find

PR #688, opened 2024-05-30
tweak: Attempt to help UI consistency

ToeKneeRED:
Leaving this as a draft until it is confirmed to help, this PR needs others to confirm: 
- Steam overlay still functions and can type in the overlay 
- Steam detects the account as in-game
- The STRUI shows more consistently than before

For me, with SKSE installed (the only way I could get the UI to consistently not show up), pressing the STRUI keys (F2/RCtrl) did nothing (across 15 game restarts). After commenting out STR loading the steamclient64 dll, the UI showed every time (across 20 game restarts). I also confirmed the above to also work without SKSE installed to make sure SKSE wasn't loading steamclient64. To ensure this isn't a random chance, others need to confirm this.

There may have been two instances of steamclient64 being loaded (or attempting to load), the game exe should load the steamclient64 dll itself rather than STR. As far as I can tell, Steam still loads fine with the in-game overlay, and Steam detecting the account as in-game functioning as expected. As stated before, others need to confirm this.

### DIFF
--- a/Code/immersive_launcher/steam/SteamLoader.cpp
+++ b/Code/immersive_launcher/steam/SteamLoader.cpp
@@ -33,6 +33,7 @@ void Load(const fs::path& aGameDir)
     AddDllDirectory(steamPath.c_str());
 
     fs::path clientPath = steamPath / kSteamDllName;
-    LoadLibraryW(clientPath.c_str());
+    // game exe should load steam itself
+    //LoadLibraryW(clientPath.c_str());
 }
 } // namespace steam


rfortier:
Updated NVIDIA to the 2024-06027 version 656.12, ran into black screen bug. But it was game night, so rolled back (to June 4) and planned to come back and help with debugging.

Today, reloaded new driver to help with the cause. Except, no black screen bug today. Only real difference is I am now compiled with the mimalloc 2.1.7 fix (which also fixes the server crash). I think with that fix your load order patch may no longer be necessary. 

Try a new build today _without_ this PR and see if it works for you. Technically, only a relink should be necessary and xmake should pick up the new mimalloc, but I rebuilt everthing.

PR #697, opened 2024-07-19
Fix: rare animation related crash when entering a party leader's cell

miredirex:
From the looks of it, sometimes `PerformComplexAction` was called with `pActor == nullptr`

### DIFF
--- a/Code/client/Games/Animation.cpp
+++ b/Code/client/Games/Animation.cpp
@@ -120,7 +120,7 @@ bool ActorMediator::ForceAction(TESActionData* apAction) noexcept
     uint8_t result = 0;
 
     auto pActor = static_cast<Actor*>(apAction->actor);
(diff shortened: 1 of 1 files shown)


maximegmd (approved):

maximegmd:
This will require extensive testing, it appears the condition was inverted so the if was never executed, and for some reason animations worked...

miredirex:
Forget it, I think this PR can be closed (or needs to be reworded if you want to merge it). `pActor` [is being checked here](https://github.com/tiltedphoques/TiltedEvolution/blob/dffd156ffb060d09a192d1a55c16419597825e97/Code/client/Services/Generic/CharacterService.cpp#L1491-L1492) for `nullptr` before we arrive in this function

I'm still debugging my animation crash and it's not related to this if check

PR #699, opened 2024-08-02
Tweak: improved door sync

miredirex:


Code/client/Services/Generic/ObjectService.cpp
```diff
@@ -298,6 +299,19 @@ void ObjectService::OnActivateNotify(const NotifyActivate& acMessage) noexcept
...
     }
 
+    if (pObject->baseForm->formType == FormType::Door)
+    {
+        auto remotePreActivationState = static_cast<TESObjectREFR::OpenState>(acMessage.PreActivationOpenState);
+        TESObjectREFR::OpenState localState = pObject->GetOpenState();
+
+        if (remotePreActivationState != localState)
+        {
+            // The doors are unsynced at this point. If we'll Activate the one on our side
```
maximegmd, line 309: Can we try to reconcile pre activation states or is it not useful?

miredirex: I'm working on syncing doors/activators state right during objects assignment (e.g. when joining a server or entering a leader's cell). When this is done, this if branch will become very unlikely but still wouldn't hurt to leave it there

PR #702, opened 2024-08-15
Feat: settime chat command

ToeKneeRED:
It is about the same if not the same as the [previous PR](https://github.com/tiltedphoques/TiltedEvolution/pull/510) for it from cosi.
Adds a `/settime` chat command that can be used on any server that is not public. 

Ideally, there would be some feedback from the server that it succeeded or failed besides the time change (e.g. a localized chat message, but that turned out to be too complex for me).

EDIT:
Since admins should be the only ones allowed to use the /settime chat command, there are now two different ways of becoming an admin; either by joining the server using the admin password specified in the STServer.ini file, or using the new /AddAdmin console command.
Unfortunately, spaces in names seem to be nearly impossible to deal with without replacing the characters and re-checking for the name over and over again. For the sake of this PR, you can't /AddAdmin or /RemoveAdmin on a name with an underscore in it as that is what is used to replace since spaces are more likely to be in names rather than underscores.

Three new console commands:

/AddAdmin <username>: add the player with the username given to the adminSessions Set
ex.
/AddAdmin Toe_Knee = Toe Knee
/AddAdmin ToeKnee = ToeKnee
/RemoveAdmin <username>: remove the player with the username given from the adminSessions Set
/admins: list out all admins

The PR also now includes a fix for a crash with the ConsoleRegistry's manual resize once it reaches a size() of 10 (it tried to erase m_commandHistory.end()).

### DIFF
--- a/Code/client/Services/TransportService.h
+++ b/Code/client/Services/TransportService.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "Events/ConnectedEvent.h"
+#include "Events/DisconnectedEvent.h"
 
 #include <atomic>
 #include <Client.hpp>
@@ -39,7 +40,8 @@ struct TransportService : Client
 protected:
     // Event handlers
     void HandleUpdate(const UpdateEvent& acEvent) noexcept;
-    void HandleConnect(const ConnectedEvent& acEvent) noexcept;
+    void HandleConnected(const ConnectedEvent& acEvent) noexcept;
+    void HandleDisconnected(const DisconnectedEvent& acEvent) noexcept;
 
     // Packet handlers
     void HandleAuthenticationResponse(const AuthenticationResponse& acMessage) noexcept;
@@ -56,5 +58,6 @@ private:
     entt::scoped_connection m_sendServerMessageConnection;
     entt::scoped_connection m_settingsChangeConnection;
     entt::scoped_connection m_connectedConnection;
+    entt::scoped_connection m_disconnectedConnection;
     std::function<void(UniquePtr<ServerMessage>&)> m_messageHandlers[kServerOpcodeMax];
 };
--- a/Code/client/Services/TransportService.h
+++ b/Code/client/Services/TransportService.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "Events/ConnectedEvent.h"
+
 #include <atomic>
 #include <Client.hpp>
 
@@ -32,10 +34,12 @@ struct TransportService : Client
 
     [[nodiscard]] bool IsOnline() const noexcept { return m_connected; }
     void SetServerPassword(const std::string& acPassword) noexcept { m_serverPassword = acPassword; }
+    const uint32_t& GetLocalPlayerId() const noexcept { return m_localPlayerId; }
 
 protected:
     // Event handlers
     void HandleUpdate(const UpdateEvent& acEvent) noexcept;
+    void HandleConnect(const ConnectedEvent& acEvent) noexcept;
 
     // Packet handlers
     void HandleAuthenticationResponse(const AuthenticationResponse& acMessage) noexcept;
@@ -46,9 +50,11 @@ private:
     entt::dispatcher& m_dispatcher;
     bool m_connected;
     String m_serverPassword{};
+    uint32_t m_localPlayerId;
 
     entt::scoped_connection m_updateConnection;
     entt::scoped_connection m_sendServerMessageConnection;
     entt::scoped_connection m_settingsChangeConnection;
+    entt::scoped_connection m_connectedConnection;
     std::function<void(UniquePtr<ServerMessage>&)> m_messageHandlers[kServerOpcodeMax];
 };
--- a/Code/server/GameServer.cpp
+++ b/Code/server/GameServer.cpp
@@ -222,6 +222,14 @@ void GameServer::Kill()
     m_requestStop = true;
 }
 
+void GameServer::RemoveAdminSession(ConnectionId_t acSession) noexcept
+{
+    if (m_adminSessions.contains(acSession))
+    {
+        m_adminSessions.erase(acSession);
+    }
+}
+
 bool GameServer::CheckMoPo()
 {
     if (!bEnableModCheck)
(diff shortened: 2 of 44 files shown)


ToeKneeRED:
Since admins should be the only ones allowed to use the `/settime` chat command, there are now two different ways of becoming an admin; either by joining the server using the admin password specified in the STServer.ini file, or using the new `/AddAdmin` console command. 
Unfortunately, spaces in names seem to be nearly impossible to deal with without replacing the characters and re-checking for the name over and over again. For the sake of this PR, you can't `/AddAdmin` or `/RemoveAdmin` on a name with an underscore in it as that is what is used to replace since spaces are more likely to be in names rather than underscores.

Three new console commands: 
- `/AddAdmin <username>`: add the player with the username given to the adminSessions Set
 ex. 
`/AddAdmin Toe_Knee` = Toe Knee 
`/AddAdmin ToeKnee` = ToeKnee
- `/RemoveAdmin <username>`: remove the player with the username given from the adminSessions Set
- `/admins`: list out all admins
     
The PR also now includes a fix for a crash with the ConsoleRegistry's manual resize once it reaches a size() of 10 (it tried to erase m_commandHistory.end()).

Code/server/GameServer.cpp
```diff
@@ -383,6 +391,129 @@ void GameServer::BindServerCommands()
                 out->error("Day must be between 0 and 31, month must be between 0 and 11, and year must be between 0 and 999.");
             }
         });
+
+    m_commands.RegisterCommand<std::string>(
+        "AddAdmin", "Add admin privileges from player", [&](Console::ArgStack& aStack) {
```
Force67, line 396: to player

Code/server/GameServer.cpp
```diff
@@ -383,6 +391,129 @@ void GameServer::BindServerCommands()
...
+                    out->info("{} admin privileges added", cUsername.c_str());
+                    playerFound = true;
+                    return;
+                }
+            });
+
+            if (!playerFound)
+            {
+                // space in username handling
+                String backupUsername = cUsername;
```
Force67, line 416: Should factor out the user name sanitizing / cleaning into a separate utility fn.

Code/server/GameServer.cpp
```diff
@@ -383,6 +391,129 @@ void GameServer::BindServerCommands()
...
+    m_commands.RegisterCommand<std::string>(
+        "RemoveAdmin", "Remove admin privileges from player", [&](Console::ArgStack& aStack) {
+            auto out = spdlog::get("ConOut");
+
+            const auto& cUsername = aStack.Pop<String>();
+            bool playerFound = false;
+
+            for (const auto& cAdmin : m_adminSessions)
+            {
+                Player* pPlayer = PlayerManager::Get()->GetByConnectionId(cAdmin);
```
Force67, line 449: Probably the get user by name stuff too. maybe have a fn there to get a user by name.

Code/server/GameServer.cpp
```diff
@@ -383,6 +391,129 @@ void GameServer::BindServerCommands()
...
+                return;
+            }
+
+            String output = "Admins: ";
+            bool _first = true;
+
+            for (const auto& cAdmin : m_adminSessions)
+            {
+                const auto& cUsername = PlayerManager::Get()->GetByConnectionId(cAdmin)->GetUsername();
+
```
Force67, line 503: Even though unlikely, we should still handle if GetConnectionId somehow returns null

Code/client/Services/TransportService.h
```diff
@@ -32,10 +34,12 @@ struct TransportService : Client
 
     [[nodiscard]] bool IsOnline() const noexcept { return m_connected; }
     void SetServerPassword(const std::string& acPassword) noexcept { m_serverPassword = acPassword; }
+    const uint32_t& GetLocalPlayerId() const noexcept { return m_localPlayerId; }
 
 protected:
     // Event handlers
     void HandleUpdate(const UpdateEvent& acEvent) noexcept;
+    void HandleConnect(const ConnectedEvent& acEvent) noexcept;
```
Force67, line 42: We should for safety reasons tombstone the localplayerid on disconnect. e.g. set it to invalid.

Code/server/GameServer.cpp
```diff
@@ -747,8 +879,14 @@ void GameServer::HandleAuthenticationRequest(const ConnectionId_t aConnectionId,
     }
 
     // check if the proper server password was supplied.
-    if (acRequest->Token == sPassword.value())
+    if (acRequest->Token == sPassword.value() || (acRequest->Token == sAdminPassword.value() && !sAdminPassword.empty()))
```
Force67, line 882: might be nicer to assign the result of the second expr to a boolean then we can also use the result below

Code/server/GameServer.h
```diff
@@ -78,6 +78,14 @@ struct GameServer final : Server
     {
         return m_isPasswordProtected;
     }
+    void SetPublic(bool aIsPublic) noexcept
+    {
+        m_isPublic = aIsPublic;
+    }
+    bool IsPublic() const noexcept
```
Force67, line 85: nodiscard

Code/server/GameServer.h
```diff
@@ -78,6 +78,14 @@ struct GameServer final : Server
     {
         return m_isPasswordProtected;
     }
+    void SetPublic(bool aIsPublic) noexcept
```
Force67, line 81: There is no need for this. the value if a server is "public" is contained by bAnnounceServer convar. if you need it from another cpp file, you can extern Convar<bool> name it.

Force67 (changes requested):
Also please run clang format python script (tools/scripts/clang_format.py) over your changes.

ToeKneeRED:
clang-format'd all files except those in the scripting expansion because I accidentally rebased instead of merge and it wants to re-add those files

Code/server/GameServer.cpp
ToeKneeRED: Fixed

Code/client/Services/TransportService.h
ToeKneeRED: Fixed with new HandleDisconnected

Code/server/GameServer.cpp
ToeKneeRED: Added two things for this, GameServer::GetAdminByUsername and PlayerManager::GetByUsername
Reworked that part to adapt it in as well

Code/server/GameServer.cpp
ToeKneeRED: Now has a nullptr check and prints to console if m_adminSessions contains a session it shouldn't

Code/server/GameServer.cpp
ToeKneeRED: Fixed with additional bool above

Code/server/GameServer.h
ToeKneeRED: Removed it, and the reference to it that was in ServerListService

Force67 (approved):

rfortier:
I'm late to the party since this was just merged. But I added it to the latest Modded Animation pre-release binary build, after verifying (shocker, it just works). So we can get a few more testers hands-on before a Master release.

PR #705, opened 2024-08-21
Exception handling fix so SetUnhandledExceptionFilter works

rfortier:
Enables various crashloggers to work with STR.

CrashHandler is always constructed now, but is IS_MASTER aware and disables minidumps when IS_MASTER is true. Unless you turn the flag back on with a debugger.

### DIFF
--- a/Code/client/CrashHandler.cpp
+++ b/Code/client/CrashHandler.cpp
@@ -31,14 +31,13 @@ LONG WINAPI VectoredExceptionHandler(PEXCEPTION_POINTERS pExceptionInfo)
 
     // Check for severe, not continuable and not software-originated exception
     if (pExceptionInfo->ExceptionRecord->ExceptionCode >= 0x80000000 &&
+        pExceptionInfo->ExceptionRecord->ExceptionCode != 0xE06D7363 && // Sigh. C++ exceptions forgot the SOFTWARE_ORIGINATE flag.
         (pExceptionInfo->ExceptionRecord->ExceptionFlags & EXCEPTION_SOFTWARE_ORIGINATE) == 0 && 
         alreadyCrashed++ == 0)
     {
         spdlog::critical (__FUNCTION__ ": crash occurred!"); 
         
-        auto present = IsDebuggerPresent() ? "already" : "not";
-        spdlog::error(__FUNCTION__ ": debugger is {} present at critical exception time, code{:x}, address{}, flags {:x} ",
-                      present,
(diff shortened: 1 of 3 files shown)


rfortier:
It looks like a lot of lines changed, but many are just indents because I wrapped pieces in a try{} to add more robustness.

### Description of changes
- STR ends up disabling support for ::SetUnhandledExceptionFilter(). That ends up being used by several mods that analyze crashes in ways that are useful and complementary to STR minidumps.
- The change "does both." If STR would create a minidump before, now it creates a minidump and calls the unhandled exception filter if it looks like there is one.
- This is a workaround. Vince indicates a proper fix needs to fully support Structured Exception Handling function tables, which IMHO is too much risky work just to get a crashhandler to run and produce traces and logs.
- The CrashHandler is now always constructed, since it carries the fix. It makes a runtime decision to suppress minidumps if IS_MASTER is asserted.
- You can change a static variable to turn them back on. Might be useful with field issues.

### How it works
- Records the UnhandledExceptionFilter at construction time.
- At crash time, if enabled, VectoredExceptionHandler() generates a minidump as before.
- Unconditionally, it checks if the current UnhandledExceptionFilter matches what it originally saved. If it changed, someone changed it and presumably wants it to run, so we invoke it. That's not ideal since it will skip other potential handlers, but it generally better than not running at all; the system is crashing.
- There are other robustness fixes. A mutex for only one crash dump, this function _will_ be called many times while generating a dump, possibly with multiple access violations. A try{} on the crash dump in case it fails. A wider net cast for unrecoverable exceptions, rather than just EXCEPTION_ACCESS_VIOLATION

Code/client/CrashHandler.cpp
```diff
@@ -22,58 +22,120 @@ std::string SerializeTimePoint(const time_point& time, const std::string& format
...
-                            /*
-                            //MiniDumpWithPrivateReadWriteMemory | // this one gens bad dump
-                            MiniDumpWithUnloadedModules |
-                            MiniDumpWithFullMemoryInfo |
-                            MiniDumpWithTokenInformation |
-                            MiniDumpWithPrivateWriteCopyMemory |
-                            */
-                            0;
+    // Serialize 
+    static std::mutex singleThreaded;;
```
RobbeBryssinck, line 29: Syntax typo, double semicolon

RobbeBryssinck (changes requested):

PR #706, opened 2024-08-31
Adjust enemy stats based on player count

otsffs:
- Increase enemy health, mana, and stamina according to the number of players.


miredirex:
We shall not do any gameplay balance tweaks, certainly not in the core part of the code. Instead, you can create and expose new Papyrus functions (like `SkyrimTogetherUtils.TotalPlayerCount()`) and create a Papyrus-based mod that would do this

PR #724, opened 2024-10-16
Add custom keybinds

ToeKneeRED:
Relies on https://github.com/tiltedphoques/TiltedHooks/pull/7

NOTES:
- This could use additional testing, especially with keyboard layouts besides standard US
  - To test:
    -  Should be able to set keys via config with the below names (or single-character key name, ex. $), or use the Keybinds window in the Debug UI
    - Should be able to cancel rebinding by pressing Escape
    - Should not be able to bind both to the same key
    - Should support wide characters
- Does not support mouse buttons
- Currently only supports a single key for each bind
- Key defaults:
  - STR UI: F2
  - Debug UI: F3
- I will add a list of currently known-acceptable key names on the wiki that can be used (as far as I know) in the config beforehand so players can set the keybind via the config without having to open the rebind menu

List of known acceptable key names (_should accept any single-character key based on keyboard layout as well_): Backspace, Tab, Enter, LSHIFT, RSHIFT, LCTRL, RCTRL, LALT, RALT, Pause, Caps Lock, IME Kana mode, IME Kanji mode, Esc, IME convert, IME nonconvert, Space, Page Up, Page Down, End, Home, Arrow Left, Arrow Up, Arrow Right, Arrow Down, Ins, Del, LWIN, RWIN, Applications, Sleep, NUMPAD 0, NUMPAD 1, NUMPAD 2, NUMPAD 3, NUMPAD 4, NUMPAD 5, NUMPAD 6, NUMPAD 7, NUMPAD 8, NUMPAD 9, NUMPAD *, NUMPAD +, NUMPAD -, NUMPAD ., NUMPAD /, F1, F2, F3, F4, F5, F6, F7, F8, F9, F10, F11, F12, F13, F14, F15, Num Lock, ScrLk, Browser Back, Browser Forward, Browser Refresh, Browser Stop, Browser Search, Browser Favorites, Browser Home, Volume Mute, Volume Down, Volume Up, Next Track, Previous Track, Stop Media, Start Mail, Select Media

---
### **KeybindService**
  - Creates/loads `keybinds.ini` config file, sets keys (for KeybindService, InputService, and DebugService)
    - If the user changes a keybind via the config file, the key will not be "confirmed" until it is pressed
    - If the user did not change the keybind via the config file, it will "confirm" the keybind
  - If a key is not confirmed, it attempts to ensure the correct keycodes are tied together by checking either VirtualKey via `OnVirtualKeyPress` or DirectInput via `OnDirectInputKeyPress`
    - Checks all aspects of a Key to see if they match (name, VirtualKey, or DirectInput)
  - Rebinding a key via its debug menu:
    - Press Escape to cancel rebinding
    - Press the new key to rebind either UI or Debug to:
      - `BindDebugKey` / `BindUIKey`
        - If the keycode equals the other keybind's keycode or its own, cancel the rebind (should not be able to rebind both to the same key)

---
### **InputService**
   - Add `Toggle` to call all the necessary UI toggle code
   - Triggers new `KeyPressEvent` to send out VirtualKey as necessary
   - Add STR local UI key
 
---
### **DebugService**
   - Add `m_debugKeyPressed` to manage `m_showDebugStuff`
   - Add KeybindView
   - Add Local debug UI key


ToeKneeRED:
Not going to have enough time or interest to make this into a proper implementation

Issue #723, opened 2024-10-05
Party member with vampirism is subject to crashing when the party leader appears in their cell


miredirex:
## Version of Skyrim Together
`dev` branch, v1.6.7, v1.6.5

No additional mods, only the required ones

## Description
When a party member (guest) has a status of a vampire lord, there's a very high chance they might crash when the party leader loads into their world (basically when their NPC character appears). Even if the guest is entering cells _last_, they could still crash. The player's active form/transformation doesn't matter, crashes occur in both human and vampire lord forms

## Steps to reproduce

This can be reproduced on a fresh pair of characters right after Helgen. You can either get vampirism naturally, or via console commands. For simplicity's sake I'm going to use console commands

1. Connect to the server as character A
2. Create a party as character A
3. Connect to the server as character B
4. Make sure you're in a party
5. As character B, enter `player.setrace yourraceracevampire` (for khajiit it's `player.setrace khajiitracevampire`, for nord it's `player.setrace nordracevampire`, etc.) and `player.addspell 0200283b` to get the spell
6. Transform into a vampire lord once, then return to human form (I'm not sure whether this step can be skipped, but do it anyway)
7. Now as character B you can just stand still and do nothing for the rest of it, just make sure to be nearby
8. As character A (the party leader), start going into and out of interior cells (entering _different_ houses) until character B crashes. This can take over 10 attempts. You'll have a better chance of crashing in crowded cities like Whiterun

The most common crash sites are 0x140443C43 (stack leads back to `HookPerformAction`) and 0x140231AF8 (stack leads back to `EquipManager::UnequipAll`), but these are not the only ones

## Reproduction rate

I'd say 60-80%

- [ ] Consistent: I can reproduce this bug by following the steps to reproduce 100% of the time (*every time*)
- [x] Occasional: I can reproduce this bug by following the steps to reproduce around 50% of the time
- [ ] Rare: I can reproduce this bug by following the steps to reproduce around 10% or less of the time

## Expected result
No crash

## Workaround

Cure vampirism

miredirex:
#725 fixes 0x140443C43, #726 fixes 0x140231AF8

miredirex:
Fixed in [1.7.0](https://github.com/tiltedphoques/TiltedEvolution/pull/742)

PR #726, opened 2024-10-24
Fix cell-change crash from SetActorInventory() calling s_unequipAll followed by s_removeAllItems

rfortier:
This fixes a cell change crash provoked by bugs in Actor::SetActorInventory(), but really a bug in the downcall to EquipManager::UnequipAll() which invokes skyrim-native TUnequipAll.

SetActorInventory() effectively calls UnequipAll(), RemoveAllItems(), then sets the new inventory with AddRemoveItem(). But the RemoveAllItems() script _also_ Unequips() each item, which seems to trigger a bug because the UnequipAll is still animating unequips. 

First fix is comment out the unequipall.

Second fix is remove as many SetActorInventory() calls as possible. In particular, when you set the inventory on a local Actor, you don't also have to set it again when the change reflects back to you in OnRemoteSpawnDataReceived(). That's just a more elaborate way of provoking overlapping Unequips (basically, back-to-back RemoveAllItems())

Partially addresses https://github.com/tiltedphoques/TiltedEvolution/issues/723 (0x140443C43 crash site)

### DIFF
--- a/Code/client/Games/Skyrim/Actor.cpp
+++ b/Code/client/Games/Skyrim/Actor.cpp
@@ -395,7 +395,9 @@ void Actor::SetActorInventory(const Inventory& aInventory) noexcept
 {
     spdlog::info("Setting inventory for actor {:X}", formID);
 
-    UnEquipAll();
+    // The UnEquipAll() that used to be here is redundant,
+    // as RemoveAllItems() unequips every item if needed.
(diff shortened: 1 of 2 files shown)


rfortier:
Reverted the 2nd fix for now, as is riskier and may be supplanted by a more comprehensive fix.

PR #728, opened 2024-10-30
Change `D3D11CreateDevice` invocation in NvidiaUtil to a dynamic one

miredirex:
Black screen fix still applies. Dynamic invocation is needed in order not to link `d3d11.dll` at compile-time (compile-time linking breaks ENB and other mods that use d3d11.dll as a proxy)


maximegmd (approved):
LGTM

PR #735, opened 2024-11-12
Sync Remove Spell / Magic Cancel

rfortier:
I'm just the PR monkey, this fix is contributed by @MostExcellent.

This allows the function that explicitly removes magic effects to sync when applied to a player. The Remove Spell function is in vanilla Skyrim but won't be synced by STR without this fix. 

Updated to newest newest /dev branch, needed to move Actor::RemoveSpell to Actor.cpp

This functionality has been in the Animation branch since before the release of 1.6.7, just got the author's permission to PR/merge it.


RobbeBryssinck:
I'm not sure what this is supposed to do. Can you give an example of what bug this would fix for example? For example, are magic effects not cancelled properly without this PR, and if so, which ones?

rfortier:
> I'm not sure what this is supposed to do. Can you give an example of what bug this would fix for example? For example, are magic effects not cancelled properly without this PR, and if so, which ones?

Suspected I would get this question. But I don't know where vanilla Skyrim uses it. I guess I can try a breakpoint and see if it fires with the way I play. MCO players discovered it because some effects in MCO weren't working/syncing right, and found it was using this vanilla Skyrim function that wasn't being synced.

marceloarc:
This pr fixed an annoying bug where healing hands was not working very well, i'm sure this fixed other bug as well

Code/client/Games/Skyrim/Actor.cpp
```diff
@@ -352,6 +352,19 @@ void Actor::StopCombat() noexcept
     s_pStopCombat(this);
 }
 
+void Actor::RemoveSpell(MagicItem* apSpell) noexcept
```
RobbeBryssinck, line 355: This function will cause the `HookRemoveSpell` hook to trigger. Instead, the hook should be moved to `Actor.cpp`, and this function should call `RealRemoveSpell` instead of fetching the address of `RemoveSpell`. For an example on how to do this, check out `InitiateMountPackage`, `RealInitiateMountPackage` and `HookInitiateMountPackage`.

Code/client/Games/Skyrim/Magic/MagicTarget.cpp
```diff
@@ -181,6 +185,26 @@ bool TP_MAKE_THISCALL(HookAddTarget, MagicTarget, MagicTarget::AddTargetData& ar
     }
 }
 
+// Designed to run when we hook removespell (AddressLib ID is 38717)
+// Sends a message to the server to remove the spell from this player on other clients
+bool TP_MAKE_THISCALL(HookRemoveSpell, Actor, MagicItem* apSpell)
```
RobbeBryssinck, line 190: This should be moved to `Actor.cpp`.

Code/encoding/Messages/RemoveSpellRequest.h
```diff
@@ -0,0 +1,16 @@
...
+#include <Structs/GameId.h>
+
+struct RemoveSpellRequest final: ClientMessage
+{
+    static constexpr ClientOpcode Opcode = kRequestRemoveSpell;
+    RemoveSpellRequest() : ClientMessage(Opcode) {}
+    virtual ~RemoveSpellRequest() = default;
+    void SerializeRaw(TiltedPhoques::Buffer::Writer& aWriter) const noexcept override;
+    void DeserializeRaw(TiltedPhoques::Buffer::Reader& aReader) noexcept override;
+    bool operator==(const RemoveSpellRequest& achRhs) const noexcept { return TargetId == achRhs.TargetId && SpellId == achRhs.SpellId && Opcode == achRhs.Opcode; }
```
RobbeBryssinck, line 12: The Opcode should be compared first.

RobbeBryssinck (changes requested):

rfortier:
This is the first time I implemented an It() / HookIt() / RealIt() pattern, so I pushed the branch early for comment on whether I got all the parts in the right places.

I'll resolve the review comments after I finish testing.

rfortier:
Changes are in and tested to the extent I can. I recommend getting an MCO user to verify, or someone who knows of a reliable vanilla test.

Code/client/Games/Skyrim/Actor.cpp
```diff
@@ -352,17 +355,15 @@ void Actor::StopCombat() noexcept
...
+bool Actor::RemoveSpell(MagicItem* apSpell) noexcept
 {
-    TP_THIS_FUNCTION(TRemoveSpell, void, Actor, MagicItem*);
-    POINTER_SKYRIMSE(TRemoveSpell, removeSpell, 38717);
     if (!apSpell)
     {
-        spdlog::error("Actor::RemoveSpell: apSpell is null");
-        return;
+        spdlog::error(__FUNCTION__ ": apSpell is null");
+        return FALSE;
```
RobbeBryssinck, line 363: tiny detail, but C++ has built in bool types, stylistically I'd prefer those over the msvc typedef'd TRUE and FALSE.

rfortier: OK, I'll clean that up.

Also, found a bit more to delete in MagicTarget.cpp as cleanup.

RobbeBryssinck (approved):

PR #737, opened 2024-11-15
Fix: Resolve animal sliding issue

otsffs:
Multiple behavior replacers have the same signature, choosing the first one.


rfortier:
C/P of discord thread to keep the history:

Normally I'd comment on the PR, but there might be some general interest or maybe someone can comment on the history of critter animation vars. 

First off, your PR is fine and IMHO should be merged. That change is what I intended originally, I must have fat-fingered it after some interruption. Don't all bugs come from interruptions? The funny thing is I never saw sliding critters so never was poked to fix it; you must have a critter animation mod.

There are two potential better fixes I've considered. First, although there is no way to tell those half-dozen critters apart, weirdly some of them do sync more vars than others. I thought about hunting through the matches to try to find the largest number of vars to sync, or maybe a "best" set by some criteria. 

The proper fix I was going to do as part of the code-cleanup pass I've promised, but it's more dangerous so should wait until after this release. That one is to:
a) hunt through the full set of animation vars for all those critters and see if there is something that can distinguish them; or 
b) actually add something to the animation vars to distinguish them.

I did a quick pass at a) and didn't find anything, but it was a quick pass.

PR #748, opened 2024-11-30
Turkish Translation Update

WeirdYunus:
Added "Reveal Players"


miredirex (changes requested):
Json is invalid, you need to add commas to the preceding entries

PR #753, opened 2024-12-03
Proof-of-concept work to identify & fix sources of logging loops

rfortier:
(log messages that repeat forever).

For discussion. Would never propose merging such ugly hacks.

### DIFF
--- a/Code/client/Services/Generic/CharacterService.cpp
+++ b/Code/client/Services/Generic/CharacterService.cpp
@@ -676,11 +676,17 @@ void CharacterService::OnBeastFormChange(const BeastFormChangeEvent& acEvent) co
 
     Actor* pActor = Utils::GetByServerId<Actor>(serverId);
     if (!pActor)
+    {
+        spdlog::warn(__FUNCTION__ ": could not find actor server id {:X}", serverId);
         return;
+    }
 
     TESNPC* pNpc = Cast<TESNPC>(pActor->baseForm);
     if (!pNpc)
+    {
+        spdlog::warn(__FUNCTION__ ": could not find actor server id {:X}", serverId);
         return;
+    }
 
     pNpc->Serialize(&request.AppearanceBuffer);
     request.ChangeFlags = pNpc->GetChangeFlags();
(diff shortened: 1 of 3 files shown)


rfortier:
I run with debug console typically, and typically after a few hours of playing the logging loops are so bad the game slows down. So I went after the problem.

The bulk of the logs that loop and don't go away are caused by AppliedQueuedEffects and related routines. These "queue" effects until conditions are met like actors being fully created, or assigned a server ID. But the problem is they never give up; they try and log endlessly. If you are a magic user (I usually am), the build up over time until the game is unplayable.

Code/client/Services/Generic/MagicService.cpp
```diff
@@ -314,15 +314,15 @@ void MagicService::OnAddTargetEvent(const AddTargetEvent& acEvent) noexcept
...
+        m_queuedEffects[acEvent.TargetID] = std::pair<int, AddTargetRequest>(0, request);
         return;
     }
 
     std::optional<uint32_t> serverIdRes = Utils::GetServerId(*it);
     if (!serverIdRes.has_value())
     {
         spdlog::warn("Server id not found for magic add target, form id: {:X}", acEvent.TargetID);
-        m_queuedEffects[acEvent.TargetID] = request;
+        m_queuedEffects[acEvent.TargetID] = std::pair<int, AddTargetRequest>(0, request);
```
Force67, line 325: make_pair pls

Code/client/Services/Generic/CharacterService.cpp
```diff
@@ -675,12 +675,20 @@ void CharacterService::OnBeastFormChange(const BeastFormChangeEvent& acEvent) co
     request.ActorId = serverId;
 
     Actor* pActor = Utils::GetByServerId<Actor>(serverId);
+    {
```
maximegmd, line 678: Missing a condition?

maximegmd (changes requested):

Code/client/Services/Generic/MagicService.cpp
rfortier: It's even simpler since C++17. But I refuse to use the "simpler" but less readable initializer syntax, '= {0, request};'

PR #754, opened 2024-12-05
Animation README language tweaks now that it is released.

rfortier:
Now that the animation support is officially merged and released, the animation readme needed a few tweaks. 

Recalling my comments on earlier drafts, I did my best to anticipate language the Dev team would agree with, but let me know if you'd like any tweaks. Or rewrites.


Force67:
Official is spelled wrong

rfortier:
> Official is spelled wrong

Added commit to fix it.

PR #757, opened 2025-01-05
Fixed sliding NPCs by adding a replay cache for animations

miredirex:
Having a single `LatestAction` wasn't enough to bring Actors into a correct animation state. Now non-cell-owners (aka party members) will replay recently run animations when an Actor is spawned in their world

### To discuss
- Not sure if the addition of `AnimationEventLists` is even acceptable as it will be difficult to maintain the lists. Is there a better way? Comparing actions using `ends_with` or `starts_with` doesn't feel right either
- Yeah I went the hardcoded route. What other approaches can we try? Serializing and sending animation graphs seems infeasible
- Should the server manage all this animation logic? What could be delegated to clients?

### TODOs
- [ ] More testing: test with 3-4 players, test dead NPCs, test dragons, test draugrs stuck in sleeping positions
- [ ] `GetSitState`, `IsSwimming`, `GetSleepState` checks and state enforcing
- [ ] Sometimes `moveStart` is completely missing when it shouldn't
- [ ] Substitute normal animations for `Instant` counterparts when possible
- [ ] [Double pass of weapon drawing](https://github.com/tiltedphoques/TiltedEvolution/blob/869d34f95df18458d56fa2eb87a99f70ed93e9cc/Code/client/Services/Generic/CharacterService.cpp#L1643) sometimes cancels the latest queued animation
- [ ] Clear replay cache when a cell has no owner/when an NPC loses its owner?
- [ ] Extend F3 entity info for easier debugging
- [ ] Reorganize code a bit: ReplayCache class/struct
- [ ] Crashes found during testing: `1401909EC → 1402C3957`, `14067644e → 140654fa4 → 1402b789a`, `14067644e → 14064ff77 → 1402b789a`. Might not be related to this fix at all, I don't know. Most often occurred during reconnects on the follower's side with lots of NPCs around


miredirex:
Going to submit a new PR very soon

PR #759, opened 2025-01-07
Case-insensitive server console commands/settings

ToeKneeRED:
This was previously mentioned in the discord but server console commands should be case-insensitive. 
To preserve their capitalization, I decided against saving commands/settings as all lowercase so the `/help` command would look nicer than many completely lowercase command/setting names.
ex. 
- `/set gameserver:spassword pass`
- `/UPTIME` 

It also adds another expected option to `BoolifyString`.


Force67:
std::string data = "Abc";
std::transform(data.begin(), data.end(), data.begin(),
    [](unsigned char c){ return std::tolower(c); });

PR #769, opened 2025-02-24
Support Quest Item Synchronization and Party Member Progression

otsffs:
**1. Quest Item Synchronization:**
Fixed an issue where quest items were incorrectly removed from inventories.
Ensured quest items are identified via IsQuestItem and properly retained and synchronized across relevant actor inventories.
**2. Party Member Quest Progression Support:**
Implemented functionality allowing party members to accept quests by interacting with NPCs.
Enabled party members to obtain quest items and use them to advance quest progress.

### DIFF
--- a/Code/client/Games/Skyrim/TESObjectREFR.cpp
+++ b/Code/client/Games/Skyrim/TESObjectREFR.cpp
@@ -781,6 +781,56 @@ void TESObjectREFR::SetInventory(const Inventory& aInventory) noexcept
     }
 }
 
+Vector<uint32_t> TESObjectREFR::RemoveNoQuestItems(Inventory& aCurrentInventory) noexcept
+{
+    ScopedEquipOverride equipOverride_;
+
+    Vector<uint32_t> questEntries{};
+
+    auto& modSystem = World::Get().GetModSystem();
+
+    for (auto& entry : aCurrentInventory.Entries)
+    {
+        if (entry.IsQuestItem)
+        {
+            uint32_t gameId = modSystem.GetGameId(entry.BaseId);
+            questEntries.emplace_back(gameId);
+            continue;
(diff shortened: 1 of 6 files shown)


Code/client/Games/Skyrim/TESObjectREFR.cpp
```diff
@@ -781,6 +781,56 @@ void TESObjectREFR::SetInventory(const Inventory& aInventory) noexcept
...
+}
+
+void TESObjectREFR::SetInitQuestInventory(Inventory& aCurrentInventory, const Inventory& acSourceInventory) noexcept
+{
+    spdlog::debug("Setting inventory for {:X}", formID);
+
+    ScopedInventoryOverride _;
+
+    // Remove all non-quest items first
+    Vector<uint32_t> nonQuestItemIds = RemoveNoQuestItems(aCurrentInventory);
```
miredirex, line 818: This is `questItemIds`, `RemoveNonQuestItems` returns quest items

miredirex (commented):
Is [this TODO](https://github.com/tiltedphoques/TiltedEvolution/blob/7e09f8b410e6fc8e3ce1388d92b64ca4ef4569e2/Code/client/Games/Skyrim/TESObjectREFR.cpp#L823) still relevant?

Can you describe the quest items bug in greater detail and ways to repro it? Why is retaining quest items crucial during an inventory update? Why didn't it work before with the old code?

I would also test a scenario where a quest's current stage is "Read note", the party member loots the quest item (the note), puts it into a container, and then the leader takes the note and reads it

As for the `IsLeader()` checks, might want to ask @RobbeBryssinck why this commit 4f2edc4725650c004c9fa387ec517c5c8411adf9 was made in the first place

PR #784, opened 2025-05-31
Enable compatibility with EngineFixes

rfortier:
Presented for discussion. We want to beat it up for maybe another week before saying it's fully baked.

It took time to hunt down, but the main fix is just a few lines in ScriptExtender.cpp to load the DLLs in the right order.

The challenge is that not every fix in EF is compatible, notably the tbbmalloc issue most are aware of. The good news is every single feature of EngineFixes is individually configurable. Just turn them off and no more problem. So far, we've identified just two conflicts with STR.

To avoid people concluding EF conflicts are horrible STR bugs (I guess that was an issue in the past), the bulk of new code checks the config for a flag indicating the STR configuration has been checked. If not found it edits the config (and adds the flag). 

If you edit the config without removing the flag, you are on your own. Hopefully smart enough to find the config file is smart enough.

### DIFF
--- a/Code/immersive_launcher/stubs/DllBlocklist.cpp
+++ b/Code/immersive_launcher/stubs/DllBlocklist.cpp
@@ -1,5 +1,14 @@
+#include <filesystem>
+#include <iostream>
+#include <fstream>
+#include <regex>
+#define SPDLOG_WCHAR_FILENAMES
+#include <spdlog/formatter.h>
 
+#include "utils/Error.h"        // For Die()
 #include "DllBlocklist.h"
+#include "../Launcher.h"        // For gameDir
+#include "../TargetConfig.h"    // For PRODUCT_NAME
 
 namespace stubs
 {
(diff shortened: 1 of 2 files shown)


Dragonissa:
We already added some of the fixes in a way cleaner way. Why not move them into the core instead.

rfortier:
> We already added some of the fixes in a way cleaner way. Why not move them into the core instead.

Can you identify which ones, or have some hints about where to find them in the history? More compatibility testing / conflict identification (if any) is part of the test plan.

To answer the question, I'd just watched one contributor spend a couple of weeks porting shader fixes from EF, and Cosi started a similar, not-completed project a year or so ago. I spent less than a week of spare time to get _all_ the fixes from EF. I think this PR is, in fact, the cleaner way to get a lot of fixes. Plus, it addresses community asks for EF compatibility since some mods require it

rfortier:
There have been a few updates to the PR and description. It should be good to go now, not sure if a comment is needed to trigger.

Code/immersive_launcher/stubs/DllBlocklist.cpp
```diff
@@ -21,8 +30,199 @@ namespace stubs
...
+
+    // Read the entire config into a string buffer.
+    // If the config doesn't exist or can't be read, it may be
+    // that the mod isn't installed at all. Or we don't have 
+    // permissions, which means the mod won't be able to read 
+    // the config either and it will be handled there.
+    // Returning Accept keeps everthing quiet (not installed is OK),
+    // and if config is protected it is the mod's job to handle it 
+    std::filesystem::path configFile = aPath / aEntry.m_configLocation;
+    std::fstream file(configFile, std::ios::in);
```
Force67, line 104: Please use the wide variant. else we get unicode troubles.

Code/immersive_launcher/stubs/DllBlocklist.cpp
```diff
@@ -21,8 +30,199 @@ namespace stubs
...
+    // Rewind config stream and prepare to generate newConfig;
+    configStream.clear(); // Clear any EOF flags
+    configStream.seekg(0, std::ios::beg);
+    std::string newConfig;   
+
+    // std::regex throws exceeptions
+    try
+    {
+        // This has to be done inside the try/catch, but will be used outside at the end.
+        std::regex sig_regex(aEntry.m_sigRegex, std::regex_constants::icase);
```
Force67: This is never used?

Code/immersive_launcher/stubs/DllBlocklist.cpp
```diff
@@ -21,8 +30,199 @@ namespace stubs
...
+
+    // Have to update the config file.
+    // If the newConfig doesn't have the signature in it, insert it to the output.
+    // Then save the new configuration
+    if (!std::regex_search(newConfig, signatureRegex))
+        newConfig = aEntry.m_sigToInsert + newConfig;
+    
+    // Have to reopen the stream to truncate it to zero, this also check write permissions
+    file.close();
+    file = std::fstream(configFile, std::ios::in | std::ios::out | std::ios::trunc);
```
Force67, line 174: It would be cleaner to just open a second fstream and desruct the other with raii. E.g. one ifstream one ofstream

Code/immersive_launcher/stubs/DllBlocklist.cpp
```diff
@@ -21,8 +30,199 @@ namespace stubs
...
+        return kGreyListAccept;
+
+    // Ask user if they want to fix config, or not load the mod.
+    if (MessageBoxW(NULL, aEntry.m_prompt, PRODUCT_NAME L" Requires Configuration Changes", MB_OKCANCEL | MB_ICONWARNING) == IDCANCEL)
+        return kGreyListAbort;
+
+    // Have to update the config file.
+    // If the newConfig doesn't have the signature in it, insert it to the output.
+    // Then save the new configuration
+    if (!std::regex_search(newConfig, signatureRegex))
```
Force67, line 170: This uses the uninitialized regex

Code/immersive_launcher/stubs/DllBlocklist.cpp
```diff
@@ -21,8 +30,199 @@ namespace stubs
...
+                std::regex replaceThis(pattern, std::regex_constants::icase);
+
+                // Stepping through these two debugging lines is very helpful for finding broken expressions.
+                std::smatch matches;
+                std::regex_search(line, matches, replaceThis);
+
+                line = std::regex_replace(line, replaceThis, withThat);
+            }
+
+            newConfig += line + "\n"; // Save the possibly edited line.
```
Force67, line 146: Mixed line endings may also be problematic.

Force67 (changes requested):

Code/immersive_launcher/stubs/DllBlocklist.cpp
rfortier: ? std::filesystem::path has polymorphic constructors including wchar_t variants. aPath came from LaunchContext::gamePath which is wchar_t*.

Code/immersive_launcher/stubs/DllBlocklist.cpp
rfortier: Flubbed refactor, rename managed to miss one?

Code/immersive_launcher/stubs/DllBlocklist.cpp
rfortier: There's actually a reason it's that way. This uses the minimum required permissions. If the configuration is correct, only ios::in is needed. If we have to edit the file we want to upgrade to write permission. Also, the only C++ standard compliant way to rewrite a file is to truncate it at open (unless you want to require more permissions, like directory change). 

Sure, could use separate istream and ostream, but then need a separate variable due to type clash; why bother when fstream serves both? This isn't high-frequency code that needs optimization.

The file.close() is redundant; that can be deleted. I was going to error check it, but then realized we don't care about an error on close at this point.

Code/immersive_launcher/stubs/DllBlocklist.cpp
rfortier: See above, completing the refactor fixed it.

Code/immersive_launcher/stubs/DllBlocklist.cpp
rfortier: On windows, by default on text mode ostreams (text mode is the default open mode), \n is automatically converted to \r\n.

I think that's what "mixed" was referencing, at least.

rfortier:
I'm on the road for the next day, will submit an updated PR after I get some time to test it.

Code/immersive_launcher/stubs/DllBlocklist.cpp
rfortier: It's fixed now.

PR #788, opened 2025-06-06
Quest Sync update: Include miscellaneous tasks and more side tasks

otsffs:
After over 100 hours of testing, no major issues have been identified.
﻿
This update enables synchronization support for a wide range of side quests, including confirmed cases such as:
- "The Wolf Queen Awakens"
- "Saints and Seducers"
﻿
With this change, the vast majority of tasks can now be synchronized. Some miscellaneous tasks may involve randomized locations, but this does not affect the synchronization of their progress.

### DIFF
--- a/Code/client/Services/Generic/QuestService.cpp
+++ b/Code/client/Services/Generic/QuestService.cpp
@@ -168,12 +168,34 @@ bool QuestService::StopQuest(uint32_t aformId)
     return false;
 }
 
+
+static const std::unordered_set<uint32_t> nonSyncableQuestIds = {
+    0x2BA16,  // Werewolf transformation quest
+    0x20071D0,  // Vampire transformation quest
+    0x3AC44,  // Unknown internal quest
+    0xFE014801,  // Unknown internal quest
+    0xF2593   // Skill experience quest
+};
+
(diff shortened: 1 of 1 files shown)


Code/client/Services/Generic/QuestService.cpp
```diff
@@ -168,12 +168,34 @@ bool QuestService::StopQuest(uint32_t aformId)
     return false;
 }
 
+
+static const std::unordered_set<uint32_t> nonSyncableQuestIds = {
```
Force67, line 172: unordered_set is overkill. lets go with an std::array, we can even make that constexpr.

Force67 (changes requested):

Code/client/Services/Generic/QuestService.cpp
rfortier: I think the concern was we don't yet know how many quests need special handling.

Force67: Thats not what i mean. Since the dev is reponsible for adding quest ids here, a set doesnt make sense if it is const. use std::array, with constexpr. ahs use the kSomethingSomethingGlobalConst convention

Force67 (changes requested):

PR #789, opened 2025-06-10
Fix: Don't process actions when graph is not ready

MostExcellent:
AnimationSystem::Update did not check if the animation graph was ready, but removed it from the vector of pending actions anyway, resulting in dropped actions.

By checking if the graph is ready and returning early if it is not, we save those actions for when we can actually try to play them back.
While this doesn't fix magic casting animations, I am working on a fix that depends on this, as the start of casting being dropped would break subesequent actions, and, therefore, the entire animation.


MostExcellent:
Edited description for clarity and accuracy

PR #794, opened 2025-06-24
Fix faction data sync issue

otsffs:
Ensure proper sync of modified faction data


miredirex:
Yep, looks like it should be checking for `false`. Oldest blame:

https://github.com/tiltedphoques/TiltedEvolution/blame/0c3659f02a1c711a1aa8d6b49b085477ecfe0ae6/Code/server/Services/CharacterService.cpp#L692-L694

PR #796, opened 2025-06-25
Fix certain spell type checks that did not check for keywords on effects

MostExcellent:
Spells themselves typically don't have keywords, which are on their effects.
As a result, returning false if the spell has no keywords leads to the check returning false incorrectly.
Something of note: while the actual use of the checks is unchanged, this actually breaks certain spell VFX which no longer show on remote clients because the spell cast is not replicated for them. This is simply a result of the intended behavior. To sync spell VFX (for example, on self-healing) without replicating the actual spell cast itself, more work is needed.

### DIFF
--- a/Code/client/Games/Primitives.h
+++ b/Code/client/Games/Primitives.h
@@ -44,6 +44,27 @@ template <class T> struct GameArray
 
     Iterator end() { return Iterator(&data[length]); }
 
+    struct ConstIterator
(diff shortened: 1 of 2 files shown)


MostExcellent:
I added a Const Iterator to GameArray, since we aren't modifying the effects, so the changed functions could all remain const

PR #799, opened 2025-07-23
Feat/tweak quest scope plus server option n logging

rfortier:
This is an update to @otsffs PR #788, which is included with this PR and so supersedes it. 
Credit to @otsffs for miscellaneous quest sync support. His original branch was merged to preserve the credit chain.

This version of the PR adds server-side support to make syncing miscellaneous quests an optional server feature, with a default setting of false (sync is disabled). All the testing we've done shows the bugs attributed to this feature (mostly, duplication of NPCs) are in fact caused by  mods or are in the base Skyrim + STR v1.7.1. We haven't found any issues with the PR.

Still, there are known exceptions, misc. quests that should not be synced. The code suppresses sync of the ones we know about, but more may be discovered. So it is prudent to leave this as an experimental feature that players enable at their own risk.

A component had to be changed to pass the quest type to the server to support the feature flag. That caused most of the changed files in the server support commit, that and the added logging logic so if another exception does come up we can find it.

@absol89 @miredirex and others have also been involved in beating up the original PR. It's solid.

### DIFF
--- a/Code/client/Services/Generic/QuestService.cpp
+++ b/Code/client/Services/Generic/QuestService.cpp
@@ -63,9 +63,22 @@ BSTEventResult QuestService::OnEvent(const TESQuestStartStopEvent* apEvent, cons
     {
         if (IsNonSyncableQuest(pQuest))
             return BSTEventResult::kOk;
-
+     
+        if (pQuest->type == TESQuest::Type::None || pQuest->type == TESQuest::Type::Miscellaneous)
+        {
+            // Perhaps redundant, but necessary. We need the logging and
+            // the lambda coming up is queued and runs later
+            GameId Id;
+            auto& modSys = m_world.GetModSystem();
+            if (modSys.GetServerModId(pQuest->formID, Id))
+                spdlog::info(__FUNCTION__ ": queuing type none/misc quest gameId {:X} questStage {} questStatus {} questType {} formId {:X} name {}",
+                             Id.LogFormat(),  pQuest->currentStage, pQuest->IsStopped() ? RequestQuestUpdate::Stopped : RequestQuestUpdate::Started,
+                             static_cast<std::underlying_type_t<TESQuest::Type>>(pQuest->type), 
+                             pQuest->formID, pQuest->fullName.value.AsAscii());
+        }
+        
         m_world.GetRunner().Queue(
-            [&, formId = pQuest->formID, stageId = pQuest->currentStage, stopped = pQuest->IsStopped()]()
+            [&, formId = pQuest->formID, stageId = pQuest->currentStage, stopped = pQuest->IsStopped(), type = pQuest->type]()
             {
                 GameId Id;
                 auto& modSys = m_world.GetModSystem();
@@ -75,6 +88,7 @@ BSTEventResult QuestService::OnEvent(const TESQuestStartStopEvent* apEvent, cons
                     update.Id = Id;
                     update.Stage = stageId;
                     update.Status = stopped ? RequestQuestUpdate::Stopped : RequestQuestUpdate::Started;
+                    update.ClientQuestType = static_cast<std::underlying_type_t<TESQuest::Type>>(type); 
 
                     m_world.GetTransport().Send(update);
                 }
@@ -97,8 +111,22 @@ BSTEventResult QuestService::OnEvent(const TESQuestStageEvent* apEvent, const Ev
         if (IsNonSyncableQuest(pQuest))
             return BSTEventResult::kOk;
 
+        if (pQuest->type == TESQuest::Type::None || pQuest->type == TESQuest::Type::Miscellaneous)
+        {
+            // Perhaps redundant, but necessary. We need the logging and
+            // the lambda coming up is queued and runs later
+            GameId Id;
+            auto& modSys = m_world.GetModSystem();
+            if (modSys.GetServerModId(pQuest->formID, Id))
+                spdlog::info(__FUNCTION__ ": queuing type none/misc quest gameId {:X} questStage {} questStatus {} questType {} formId {:X} name {}",
+                             Id.LogFormat(), pQuest->currentStage,
+                             RequestQuestUpdate::StageUpdate,
+                             static_cast<std::underlying_type_t<TESQuest::Type>>(pQuest->type),
+                             pQuest->formID, pQuest->fullName.value.AsAscii());
+        }
+
         m_world.GetRunner().Queue(
-            [&, formId = apEvent->formId, stageId = apEvent->stageId]()
+            [&, formId = apEvent->formId, stageId = apEvent->stageId, type = pQuest->type]()
             {
                 GameId Id;
                 auto& modSys = m_world.GetModSystem();
@@ -108,6 +136,7 @@ BSTEventResult QuestService::OnEvent(const TESQuestStageEvent* apEvent, const Ev
                     update.Id = Id;
                     update.Stage = stageId;
                     update.Status = RequestQuestUpdate::StageUpdate;
+                    update.ClientQuestType = static_cast<std::underlying_type_t<TESQuest::Type>>(type);
 
                     m_world.GetTransport().Send(update);
                 }
@@ -127,6 +156,11 @@ void QuestService::OnQuestUpdate(const NotifyQuestUpdate& aUpdate) noexcept
         spdlog::error("Failed to find quest, base id: {:X}, mod id: {:X}", aUpdate.Id.BaseId, aUpdate.Id.ModId);
         return;
     }
+
+    if (pQuest->type == TESQuest::Type::None || pQuest->type == TESQuest::Type::Miscellaneous)
+        spdlog::info(__FUNCTION__ ": receiving type none/misc quest update gameId {:X} questStage {} questStatus {} questType {} formId {:X} name {}",
+                     aUpdate.Id.LogFormat(), aUpdate.Stage, aUpdate.Status,
+                     aUpdate.ClientQuestType, formId, pQuest->fullName.value.AsAscii());
 
     bool bResult = false;
     switch (aUpdate.Status)
--- a/Code/encoding/Messages/NotifyQuestUpdate.h
+++ b/Code/encoding/Messages/NotifyQuestUpdate.h
@@ -15,7 +15,7 @@ struct NotifyQuestUpdate final : ServerMessage
     void SerializeRaw(TiltedPhoques::Buffer::Writer& aWriter) const noexcept override;
     void DeserializeRaw(TiltedPhoques::Buffer::Reader& aReader) noexcept override;
 
-    bool operator==(const NotifyQuestUpdate& acRhs) const noexcept { return GetOpcode() == acRhs.GetOpcode() && Id == acRhs.Id && Stage == acRhs.Stage && Status == acRhs.Stage; }
+    bool operator==(const NotifyQuestUpdate& acRhs) const noexcept { return GetOpcode() == acRhs.GetOpcode() && Id == acRhs.Id && Stage == acRhs.Stage && Status == acRhs.Stage && ClientQuestType == acRhs.ClientQuestType; }
 
     enum StatusCode : uint8_t
     {
@@ -27,4 +27,5 @@ struct NotifyQuestUpdate final : ServerMessage
     GameId Id;
     uint16_t Stage;
     uint8_t Status;
+    uint8_t ClientQuestType;
 };
--- a/Code/server/Services/QuestService.cpp
+++ b/Code/server/Services/QuestService.cpp
@@ -7,6 +7,13 @@
 #include <Messages/RequestQuestUpdate.h>
 #include <Messages/NotifyQuestUpdate.h>
 
+#include <Setting.h>
+namespace
+{
+Console::Setting bEnableMiscQuestSync{"Gameplay:bEnableMiscQuestSync", "(Experimental) Syncs miscellaneous quests when possible", false};
+
+}
+
 QuestService::QuestService(World& aWorld, entt::dispatcher& aDispatcher)
     : m_world(aWorld)
 {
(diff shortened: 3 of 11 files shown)


rfortier:
Silly Linux compile issue is fixed, should be stable now.

Code/client/Services/Generic/QuestService.cpp
```diff
@@ -97,8 +111,22 @@ BSTEventResult QuestService::OnEvent(const TESQuestStageEvent* apEvent, const Ev
         if (IsNonSyncableQuest(pQuest))
             return BSTEventResult::kOk;
 
+        if (pQuest->type == TESQuest::Type::None || pQuest->type == TESQuest::Type::Miscellaneous)
+        {
+            // Perhaps redundant, but necessary. We need the logging and
+            // the lambda coming up is queued and runs later
+            GameId Id;
+            auto& modSys = m_world.GetModSystem();
+            if (modSys.GetServerModId(pQuest->formID, Id))
```
Force67, line 122: Coding conventions dictate braces around ifs

Code/client/Services/Generic/QuestService.cpp
```diff
@@ -168,12 +202,34 @@ bool QuestService::StopQuest(uint32_t aformId)
     return false;
 }
 
+
+static const std::unordered_set<uint32_t> nonSyncableQuestIds = {
```
Force67: Please, array

Code/client/Services/Generic/QuestService.cpp
```diff
@@ -63,9 +63,22 @@ BSTEventResult QuestService::OnEvent(const TESQuestStartStopEvent* apEvent, cons
...
             return BSTEventResult::kOk;
-
+     
+        if (pQuest->type == TESQuest::Type::None || pQuest->type == TESQuest::Type::Miscellaneous)
+        {
+            // Perhaps redundant, but necessary. We need the logging and
+            // the lambda coming up is queued and runs later
+            GameId Id;
+            auto& modSys = m_world.GetModSystem();
+            if (modSys.GetServerModId(pQuest->formID, Id))
```
Force67, line 73: braces after if

Code/client/Services/Generic/QuestService.cpp
rfortier: Done

Code/client/Services/Generic/QuestService.cpp
rfortier: Done

rfortier: Done

Code/client/Services/Generic/QuestService.cpp
rfortier: Done

rfortier:
Note recommendation to treat PR #805 as a prerequisite.

Code/encoding/Messages/NotifyQuestUpdate.h
```diff
@@ -15,7 +15,7 @@ struct NotifyQuestUpdate final : ServerMessage
     void SerializeRaw(TiltedPhoques::Buffer::Writer& aWriter) const noexcept override;
     void DeserializeRaw(TiltedPhoques::Buffer::Reader& aReader) noexcept override;
 
-    bool operator==(const NotifyQuestUpdate& acRhs) const noexcept { return GetOpcode() == acRhs.GetOpcode() && Id == acRhs.Id && Stage == acRhs.Stage && Status == acRhs.Stage; }
+    bool operator==(const NotifyQuestUpdate& acRhs) const noexcept { return GetOpcode() == acRhs.GetOpcode() && Id == acRhs.Id && Stage == acRhs.Stage && Status == acRhs.Stage && ClientQuestType == acRhs.ClientQuestType; }
```
miredirex:
```suggestion
    bool operator==(const NotifyQuestUpdate& acRhs) const noexcept { return GetOpcode() == acRhs.GetOpcode() && Id == acRhs.Id && Stage == acRhs.Stage && Status == acRhs.Status && ClientQuestType == acRhs.ClientQuestType; }
```

Uh-oh, looks like this bug has been here since forever. `Status` should be compared against `Status`

And in `RequestQuestUpdate.h` too

Code/client/Services/Generic/QuestService.cpp
```diff
@@ -168,12 +208,34 @@ bool QuestService::StopQuest(uint32_t aformId)
...
+    // Unknown internal quests causing excessive logging: 0x3AC44, 0xFE014801, 0xF2593
+
+    const auto& stages = apQuest->stages;
+
+    if (std::find(nonSyncableQuestIds.begin(), nonSyncableQuestIds.end(), apQuest->formID) != nonSyncableQuestIds.end())
+    {
+        // For specific internal quests, check if they should be blocked from syncing based on their type or empty stages.
+        return apQuest->type == TESQuest::Type::None // Internal event
+               || apQuest->type == TESQuest::Type::Miscellaneous
+               || stages.Empty();
```
miredirex: Why not just "return true" if we've found a predefined non syncable quest? Aren't all of those already either None or Misc? 

Also, as a side question: why sync the None type, are there known quests that fall under this type?

Code/client/Services/Generic/QuestService.cpp
```diff
@@ -168,12 +208,34 @@ bool QuestService::StopQuest(uint32_t aformId)
     return false;
 }
 
+
+static const TiltedPhoques::Vector<uint32_t> nonSyncableQuestIds = {
```
miredirex: Didn't Vince suggest using a constexpr array? And then `std::begin` and `std::end` to iterate

Code/server/Services/QuestService.cpp
```diff
@@ -63,19 +84,15 @@ void QuestService::OnQuestChanges(const PacketEvent<RequestQuestUpdate>& acMessa
...
-            spdlog::warn("Unable to delete quest object {:X}:{:X}", message.Id.ModId, message.Id.BaseId);
+            spdlog::warn("Unable to delete quest object {:X}", message.Id.LogFormat(), message.Id.BaseId);
 
         notify.Status = NotifyQuestUpdate::Stopped;
     }
 
-    const auto& partyComponent = acMessage.pPlayer->GetParty();
-    if (!partyComponent.JoinedPartyId.has_value())
-        return;
-
```
miredirex, line 79: Are we sure that moving this if+return is safe? See `auto& newQuest = entries.emplace_back();` and the surrounding code. It may be important to sync the quest log even when not in a party, but I don't know for sure

Code/encoding/Messages/NotifyQuestUpdate.h
rfortier: Done. But this should probably also be a separate PR since the changes are important (potential non-progression of quests), and this more complicated PR may not be merged soon.

Code/client/Services/Generic/QuestService.cpp
rfortier: Yeah, I asked that the first several times I stared at that.  It has to do with those are NonSyncable only if they also meet one of the other conditions.  The most obvious case is the logic would otherwise chose to sync misc. quests, but for these, don't. 

Implied also is that the result of the if is not constant for every use of that QuestID. Dunno if that's the case, but that would also explain the conditions.

@otsffs ?

Code/client/Services/Generic/QuestService.cpp
rfortier: Vince objected to the original std::unordered_set<> as excessive for a 5 element array and suggested an array. I used Vector because TiltedPhoques::Vector exists. 

I concur, for a 5 element array, this isn't worth any time to optimize.

Code/server/Services/QuestService.cpp
rfortier: As sure as can be. In both positions it prevents any message/sync going to party. With it at the end, the code puts a log of effort into constructing a message to send, only to throw it away. If they join a party they end up coming back to this code and running it again. And this code has been in the fork for 6 months with no issues seen.

miredirex: The side effect that was previously run with if+return being in the end:

https://github.com/tiltedphoques/TiltedEvolution/blob/0f03f9c6d6346ce4f7038b891259f5c08ef6cffe/Code/server/Services/QuestService.cpp#L33-L40

It may be critical to update the player's quest log even when **not in party**. I feel like not even the dev team would remember this particular logic, so I would rather not touch this bit

rfortier: Now on a screen bigger than 5", yes, there was a side effect in the original location, updating the quest local component. I think I moved it to limit log spam, but put it back in the latest commit.

Needs some soak time, but since it is restoring original behavior I expect it to be quick.

Code/client/Services/Generic/QuestService.cpp
otsffs: This piece of code was left over from when I was debugging task types, so it can actually just return the result directly. Quests of type "None" do need to be synchronized, as many important quests are of this type—I've already provided examples in my PR.

Code/client/Services/Generic/QuestService.cpp
```diff
@@ -168,12 +208,29 @@ bool QuestService::StopQuest(uint32_t aformId)
...
+    0x2BA16,  // Werewolf transformation quest
+    0x20071D0,  // Vampire transformation quest
+    0x3AC44,  // Unknown internal quest
+    0xFE014801,  // Unknown internal quest
```
miredirex: 0x3AC44 is "MS13BleakFallsBarrowLeverScene", 0xFE014801 is [a dynamic id](https://en.m.uesp.net/wiki/Skyrim:Form_ID#Creation_Club), is it not? There is no point in keeping it here because the actual id will be different for every player; cc @otsffs

Code/client/Services/Generic/QuestService.cpp
```diff
@@ -168,12 +208,29 @@ bool QuestService::StopQuest(uint32_t aformId)
...
-    auto& stages = apQuest->stages;
-    return apQuest->type == TESQuest::Type::None // internal event
-           || apQuest->type == TESQuest::Type::Miscellaneous || stages.Empty();
+    // Quest type None and quests with no quest stages are never synced.
+    // Most type Miscellaneous quests can sync, but there's a list that must be excluded.
+    // Internal quest IDs: Werewolf transformation quest: 0x2BA16, Vampire transformation quest: 0x20071D0,
+    // Unknown internal quests causing excessive logging: 0x3AC44, 0xFE014801, 0xF2593
+
+    bool bNonSyncable = apQuest->stages.Empty();
+    if (!bNonSyncable && apQuest->type == TESQuest::Type::Miscellaneous)
```
miredirex: Needs a check for `None`. If you look up in SSEEdit, Werewolf and Vampire transformation quests are both of type `None`

Libraries/TiltedReverse
miredirex, line 1: This submodule update probably shouldn't be a part of this

miredirex (changes requested):

Code/client/Services/Generic/QuestService.cpp
rfortier: Ahh. Sounds like the correct fix then is to remove the qualifier; if it is on the nonsyncable list, it is nonsyncable.

Code/client/Services/Generic/QuestService.cpp
rfortier: Several comments swirl in this area, I've folded them together to resolve.

Code/client/Services/Generic/QuestService.cpp
rfortier: Several comments swirl in this area, I've folded them together to resolve.

Code/client/Services/Generic/QuestService.cpp
rfortier: Several comments swirl in this area, I've folded them together to resolve.

rfortier:
All PR comments incorporated. I'll do a build and shake this down, the reconvert the PR to submitted.

Libraries/TiltedReverse
rfortier: Fixed.

Code/client/Services/Generic/QuestService.cpp
```diff
@@ -168,12 +208,21 @@ bool QuestService::StopQuest(uint32_t aformId)
...
 
+static constexpr std::array<uint32_t, 4> kNonSyncableQuestIds = {
+    0x2BA16,   // Werewolf transformation quest
+    0x20071D0, // Vampire transformation quest
+    0x3AC44,   // MS13BleakFallsBarrowLeverScene
+    // 0xFE014801,  // Unknown dynamic ID, kept as note, maybe lookup correct ID this game?
+    0xF2593 // Skill experience quest
+};
+
 bool QuestService::IsNonSyncableQuest(TESQuest* apQuest)
```
rfortier: @Force67 Vince, I think I accidentally stomped your comment with a push --force.

I'd originally made it a vector since a TP version of that exists, but if the goal was constexpr, sure here's a constexpr std::array<uint32_t, 4> version of it.

Code/client/Services/Generic/QuestService.cpp
```diff
@@ -168,12 +208,21 @@ bool QuestService::StopQuest(uint32_t aformId)
     return false;
 }
 
+static constexpr std::array<uint32_t, 4> kNonSyncableQuestIds = {
+    0x2BA16,   // Werewolf transformation quest
+    0x20071D0, // Vampire transformation quest
+    0x3AC44,   // MS13BleakFallsBarrowLeverScene
+    // 0xFE014801,  // Unknown dynamic ID, kept as note, maybe lookup correct ID this game?
+    0xF2593 // Skill experience quest
+};
```
miredirex:
```suggestion
static constexpr std::array kNonSyncableQuestIds = std::to_array<uint32_t>({
    0x2BA16,   // Werewolf transformation quest
    0x20071D0, // Vampire transformation quest
    0x3AC44,   // MS13BleakFallsBarrowLeverScene
    // 0xFE014801,  // Unknown dynamic ID, kept as note, maybe lookup correct ID this game?
    0xF2593 // Skill experience quest
});
```
This avoids hardcoding the size

PR #800, opened 2025-07-24
Fix PR #796, needs exception for Bow of Shadows.

rfortier:


Code/client/Games/Skyrim/Forms/MagicItem.cpp
```diff
@@ -20,11 +20,23 @@ bool MagicItem::IsWardSpell() const noexcept
...
+    // PR #796 and the "SkyPatcher Keyword Framework" mod both hoist
+    // some keywords from the effects up to the spell, because some 
+    // spell records are buggy; this fixes things like syncing more ward spells.
+    // But note sometimes that breaks other things. This is the first one found,
+    // Bow of Shadows Invisibility (formID 0xFE003805), which doesn't have an 
+    // Invisibility keyword. Apparently deliberately, it isn't really a spell but a 
+    // condition (readying the bow) that triggers an effect. If it is treated as a 
+    // spell, syncing doesn't work. There may be others, if so, there will be a 
+    // more general fix.
+    if (formID == 0xfe003805)
```
miredirex, line 32: You can't reference dynamic form IDs in code and expect it to work, can you? They may change depending on load order https://en.m.uesp.net/wiki/Skyrim:Form_ID#Creation_Club

rfortier: I believe you are right, I thought it was fixed for CC but it is not.

Does STR have the equivalent of commonlib RE::TESForm::LookupByEditorID("ccBGSSSE038_InvisibilitySPELL")?

rfortier: I need some thoughts as to why this doesn't work. GetFormEditorID always returns an empty string.
(verbose code was so I could follow it in assembly and try to figure out what's going on. And getting around const)

```
auto pForm = const_cast<TESForm *>(dynamic_cast<const TESForm*>(this));
auto pEditorID = pForm->GetFormEditorID();
if (!std::strcmp(pEditorID, "ccBGSSSE038_InvisibilitySPELL"))
    return false;
```

rfortier: Answering my own question. It doesn't work because GetFormEditorID() is not implemented in _any_ of the various RE:: libraries, except po3's version of CommonLibSSE. Same for the inverse function, call it LookupFormByEditorID().

Now it's complicated. We're not ready to embark on embedding CommonLib. I asked ChatGPT to generate the code for the simple case, where I know the plugin file name. Not that I trust chatGPT code, I just wanted an idea of what it would take. It's a page of code, maybe two when you fill in the missing stuff.

That's a lot for a one-off exception cosmetic fix. Unless we think it will have other uses. Or if someone has a better idea for  an approach. I read through ModSystem::, it doesn't know enough to do this either. 

I'm open to suggestions. Including "just fix it in your version, no one else cares."

rfortier: I pushed an update that moves this PR from "only works with one load order" to "works with all load orders, but statistically rarely will not sync an Invisibility spell if it has the same local ID in a different ESL mod."

That's still not good enough to merge though. I'll close this PR, we're looking at implementing GetFormByEditorID() properly, since it will be useful elsewhere.

rfortier:
Closing this till we can do it properly.

PR #802, opened 2025-07-27
Fix/excess logging bugs, resubmitted

rfortier:
Resubmitting. This fixes infinite logging and ultimately huge logs and game slowing to a crawl when spells and effects are sent to the server before an Actor is fully created.

Merge readiness: good for review, a fair amount of testing done. The logs might want some adjustments before merge, they are Warnings to make them stand out in the sea of logs, can be turned down after some more mileage (maybe to debug level).

The old code captured just the last, most recent effect, threw the rest away, and then retried forever to apply the effect. Except sometimes that Actor (server or local) will never exist; that happens in fast windows like ownership changes. Mostly on cell changes, some on spawning.

First fix timed out those dead entries.

New fix implements a proper queue per Actor to maintain all the effects, still with timeouts for when it will never succeed. The logging shows a dozen or more effects isn't unusual, so this should be a big help.

### DIFF
--- a/Code/client/Services/Generic/MagicService.cpp
+++ b/Code/client/Services/Generic/MagicService.cpp
@@ -296,33 +296,35 @@ void MagicService::OnAddTargetEvent(const AddTargetEvent& acEvent) noexcept
 
     if (!m_world.GetModSystem().GetServerModId(acEvent.SpellID, request.SpellId.ModId, request.SpellId.BaseId))
     {
-        spdlog::error("{}: Could not find spell with form {:X}", __FUNCTION__, acEvent.SpellID);
+        spdlog::error("{}: could not find server ID for spell with formID {:X}, discarding", __FUNCTION__, acEvent.SpellID);
         return;
     }
 
     if (!m_world.GetModSystem().GetServerModId(acEvent.EffectID, request.EffectId.ModId, request.EffectId.BaseId))
     {
-        spdlog::error("{}: Could not find effect with form {:X}", __FUNCTION__, acEvent.EffectID);
+        spdlog::error("{}: could not find server ID for effect with formID {:X}, discarding", __FUNCTION__, acEvent.EffectID);
         return;
     }
 
     request.Magnitude = acEvent.Magnitude;
 
+    // Because it takes time to create Actors, the Caster or Target may not
+    // exist on the server yet, or server may not have told us yet. Have to queue to compensate.
     auto view = m_world.view<FormIdComponent>();
     const auto it = std::find_if(std::begin(view), std::end(view), [id = acEvent.TargetID, view](auto entity) { return view.get<FormIdComponent>(entity).Id == id; });
 
     if (it == std::end(view))
     {
-        spdlog::warn("Form id not found for magic add target, form id: {:X}", acEvent.TargetID);
-        m_queuedEffects[acEvent.TargetID] = request;
+        spdlog::warn("{}: server entity for target formID not found, formID: {:X}, queueing", __FUNCTION__, acEvent.TargetID);
+        m_queuedEffects.push(MagicAddTargetEventQueue(acEvent));
         return;
     }
 
     std::optional<uint32_t> serverIdRes = Utils::GetServerId(*it);
     if (!serverIdRes.has_value())
     {
-        spdlog::warn("Server id not found for magic add target, form id: {:X}", acEvent.TargetID);
-        m_queuedEffects[acEvent.TargetID] = request;
+        spdlog::warn("{}: server ID for target formID not found, formID: {:X}, queueing", __FUNCTION__, acEvent.TargetID);
+        m_queuedEffects.push(MagicAddTargetEventQueue(acEvent));
         return;
     }
 
@@ -332,8 +334,8 @@ void MagicService::OnAddTargetEvent(const AddTargetEvent& acEvent) noexcept
 
     if (casterIt == std::end(view))
     {
-        spdlog::warn("Form id not found for magic add target, form id: {:X}", acEvent.CasterID);
-        m_queuedEffects[acEvent.TargetID] = request;
+        spdlog::warn("{}: server entity for caster formID not found, formID: {:X}, queueing", __FUNCTION__, acEvent.CasterID);
+        m_queuedEffects.push(MagicAddTargetEventQueue(acEvent));  
         return;
     }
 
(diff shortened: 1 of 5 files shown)


Force67:
Perhaps we could also move some of these things into their own logger, so they can be turned on/off at will?

rfortier:
OK, pushed those logging changes and fixed an edge case. Logs at DEBUG level in the PR copy.

We'll shake it out for a few days / a week, but it should be good to go.

Code/client/Services/MagicService.h
```diff
@@ -115,4 +110,64 @@ struct MagicService
...
+    };
+
+  private:
+    std::queue<MagicAddTargetEventQueue> m_queuedEffects;
+    std::queue<MagicNotifyAddTargetQueue> m_queuedRemoteEffects;
 };
+
+// Exposed so we can increase log level of just this tricky code, in debugger or a build.
+// Non-const has to be initialized outside of class.
+spdlog::level::level_enum MagicService::MagicQueue::m_logLevel{spdlog::level::debug};
```
Force67, line 173: lets make this a constexpr static member variable of the magic service

Code/client/Services/MagicService.h
```diff
@@ -115,4 +110,64 @@ struct MagicService
...
+    class MagicQueue
+    {
+      public:
+        MagicQueue() = default;
+        MagicQueue(const MagicQueue&) = default;
+        ~MagicQueue() noexcept = default;
+
+        bool Expired() const noexcept { return std::chrono::steady_clock::now() > m_expiration; }
+        static spdlog::level::level_enum m_logLevel; // Initializer at file end, non-const so debugger can change.
+        template <typename... Args> static inline void spdlog(spdlog::format_string_t<Args...> aFmt, Args&&... args)
```
Force67: lets name it something more concrete like "Log"

Code/client/Services/Generic/MagicService.cpp
```diff
@@ -494,47 +511,76 @@ void MagicService::ApplyQueuedEffects() noexcept
...
+        {
+            // Process queued target events. If caster and target now have server IDs, this should work.
+            // If they don't, stop processing the list until next iteration.
+            auto view = m_world.view<FormIdComponent>();
+            const auto it = std::find_if(std::begin(view), std::end(view), [id = target.TargetID, view](auto entity) { return view.get<FormIdComponent>(entity).Id == id; });
 
-        request.TargetId = serverIdRes.value();
+            if (it == std::end(view))
+            {
+                spdlog::debug("{}: server entity for AddTargetEvent target formID still not found: caster {}({:X}), spell {:X}, effect {:X}, target {}({:X})",
```
Force67, line 536: is this supposed to be just spdlog, or MagicQueue::spdlog? also will this debug get not printed on release builds?

Code/client/Services/Generic/MagicService.cpp
```diff
@@ -494,47 +511,76 @@ void MagicService::ApplyQueuedEffects() noexcept
...
+            if (!serverIdRes.has_value())
+            {
+                spdlog::debug("{}: serverID for AddTargetEvent target formID still not found: caster {}({:X}), spell {:X}, effect {:X}, target {}({:X})",
+                              __FUNCTION__, pCasterName, target.CasterID, target.SpellID, target.EffectID, pTargetName, target.TargetID);
+                break;
+            }
 
-        markedForRemoval.push_back(formId);
+            // At this point, it will succeed or fail, but not queue another one ad infinitum
+            MagicQueue::spdlog("{}: retrying AddTargetEvent for caster {}({:X}), spell {:X}, effect {:X}, target {}({:X})",
```
Force67: same here

Force67 (changes requested):

Code/client/Services/MagicService.h
rfortier: It can't be. If you do that you can't change it in the debugger, it must be non-const. And definitely not constexpr.

I did try const at first, it's neater and I expected the debugger to be lenient. But it is not, and can't be convinced to let you change the value by getting rid of const.

Code/client/Services/MagicService.h
rfortier: I do not understand why we'd want to hide the fact it is spdlog. People have to know how to use it.

Code/client/Services/Generic/MagicService.cpp
rfortier: It's supposed to be explicitly debug level. It's _way_ to noisy to leave it on in any situation other than you need every log for some reason.

Code/client/Services/MagicService.h
Force67: Its against the naming convention to name a member fn lowercase.

rfortier: That's true.
I thought it was more important to mimic spdlog, but it can be capitalized.
I should do a general scrub for the conventions, I probably missed some.

rfortier: OK, made that change and fixed one naming inconsistency. Which identified another debatable point. Pushed so you can see it, but won't actually get play testing until tonight. It "should be" equivalent, lol. 

The debate point: I spotted one other minor break in the conventions, a member variable named m_CamelCase instead of m_camelCase. I think it / they are proper as they are, though, because they are naming a proper noun, like m_NotifyAddTarget.

rfortier:
Re-pushed to delete the file accidentally pushed.

rfortier:
One more, maybe the final comment. Several people have now run the newest and report that the recursion case is fixed. All of the reproducers have been tried and they all work now.

Good to go?

rfortier:
> Last nitpick, the **FUNCTION** prints can be simplified, by just putting them directly into the string instead of formatting them in. See: void DoStuff() { std::printf(**FUNCTION** " what?"); } this compiles on godbolt with msvc

I think formatting got you in your example, but I think I get the point.  `__FUNCTION__` interpolation.

That's true, and I generally have used the string-splicing form. Up until I recently when I found out on a different PR #799 that the linux/GCC compiler doesn't implement the `__FUNCTION__` splicing properly (it doesn't reduce to a basic char* type, so it won't splice). So for Linux code (server, common...) you have to format it in and pass it as a parameter.

That makes the string-format & pass parameter the more "portable" way, and required for code that will be compiled by Linux. Reality is `__FUNCTION__ `isn't part of the standard, just commonly supported, so there are nuances between the compilers.

Bottom line: even though MagicService is client-side and _could_ splice, I've taken to following the convention I find in the file I'm working on (which is usually best practice, though some will debate it). In that file the existing code that used `__FUNCTION__` passed it as a parameter, so I followed that pattern. I did add a lot of instances of identifying the function, that is so useful I think the change should be made. I'd go so far as to say that should be in the style guide for most (non-user-facing) things.

Bottom line: I could change it, in some of the files. Is it worth it? It's not like it could be made consistent throughout.

PR #803, opened 2025-07-29
feat: Enable compatibility with SkyrimSoulsRE to fix menu-related crashes

otsffs:
Enable compatibility with the SkyrimSoulsRE mod. If the SkyrimSoulsRE mod is not installed, menu pausing will be managed by SkyrimTogetherReborn. If SkyrimSoulsRE is installed, it will take over the management of menu pausing.

Currently, non-paused menus in SkyrimTogetherReborn are causing several crashes. Known crashes include:

1. 100% crash when binding potions to the same hotkey in the Preferences menu.
2. High probability of crashing when swapping equipment or items with a follower.

Integrating the SkyrimSoulsRE mod resolves these two specific issues and may fix additional menu-related crashes. If the SkyrimSoulsRE mod's configuration is set to match the current list of non-paused menus in SkyrimTogetherReborn, this change should theoretically introduce no additional risks.

### DIFF
--- a/Code/client/Games/Skyrim/Interface/UI.cpp
+++ b/Code/client/Games/Skyrim/Interface/UI.cpp
@@ -2,6 +2,7 @@
 #include <Games/Skyrim/Interface/UI.h>
 #include <Misc/BSFixedString.h>
 #include <TiltedOnlinePCH.h>
+#include "immersive_launcher/stubs/DllBlocklist.h"
 
 #include <World.h>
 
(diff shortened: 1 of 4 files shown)


Dragonissa:
Most of this is already implemented natively. It would rather make sense fixing the issue in code.

miredirex:
I changed my mind, I won't be looking into this, nor porting SkyrimSouls' code to STR. I just checked, Skyrim Souls RE is over 3k lines of C++. Why not delegate certain functionality to a mod that simply does it better? There are so many other things that need fixing, and this isn't one of them

PR #804, opened 2025-07-30
Add in-game message box that warns about launching from the wrong exe

miredirex:
If the user has SkyrimTogether.esp but is launching the mod using the wrong exe, then the dialog will be shown on every save load, like the SkyUI's one when you don't have SKSE installed

Thoughts on the wording?


miredirex:
Yes, the popup works since unimplemented/unbound native functions return default values, bool = false, int = 0, etc. That's the idea, basically

miredirex:
Upd. Changed `IsSkyrimTogetherExeRunning` to `DidLaunchSkyrimTogether` for clarity, and because naming is hard. Also added a brief explanation comment to the papyrus script https://github.com/tiltedphoques/TiltedEvolution/commit/57992d4a125723cc3a66ba7dc1bb480faf16d366

SkyrimTogetherUtils remains unchanged, I never touched it, but since Git detected a change I decided to commit it as well (I assume it is because of the SkyrimTogether.esp changes)

PR #805, opened 2025-08-15
Prevent multiple triggers of quest stage change

rfortier:
This started as a bug hunt for misc. quest syncing, but it effects ALL quest type progression. When the quest stage is set, syncing will cause each player to try to set it, which can cause duplication of steps or unwanted progression.

The fix is to make sure desired quest stage != current quest stage.


rfortier:
This fixes, for example, the early trigger of The Way Of The Voice when the party is returning from killing the dragon.

With a bit of luck it may fix the High Hrothgar summit (truce) quest stage which has always been hyper-buggy.

PR #808, opened 2025-09-10
Fix PR #796, needs exception for Bow of Shadows.

rfortier:
At long last, here is the real, supportable, always-works version of  the Bow of Shadows invisibility sync fix.

Credit to @miredirex for pointing me in the right direction for dealing with shifting mod IDs / formIDs so it will always work.

Deals with the one special case that BoS needs to be exempted from PR #796. If we find more it needs an array of baseFormIds and mod string structures and a loop to handle them all, but the pattern should be good.

### DIFF
--- a/Code/client/Games/Skyrim/Forms/MagicItem.cpp
+++ b/Code/client/Games/Skyrim/Forms/MagicItem.cpp
@@ -1,4 +1,5 @@
 #include "MagicItem.h"
+#include <Games/TES.h>
 
 bool MagicItem::IsWardSpell() const noexcept
 {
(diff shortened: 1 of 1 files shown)


Code/client/Games/Skyrim/Forms/MagicItem.cpp
```diff
@@ -20,11 +21,35 @@ bool MagicItem::IsWardSpell() const noexcept
...
+    // PR #796 and the "SkyPatcher Keyword Framework" mod both hoist
+    // some keywords from the effects up to the spell, because some 
+    // spell records are buggy; this fixes things like syncing more ward spells.
+    // But the fix can break other things. This is the first one found,
+    // Bow of Shadows Invisibility (formID 0xFEXXX805), which doesn't have an 
+    // Invisibility keyword. Apparently deliberately, it isn't really a spell but a 
+    // condition (readying the bow) that triggers an effect. If it is treated as a 
+    // spell, syncing doesn't work. There may be others, if so, there will be a 
+    // more general fix but this is the pattern, just needs an array / loop.
+    static bool runonce = true;
```
Force67, line 33: std::call_once

Code/client/Games/Skyrim/Forms/MagicItem.cpp
```diff
@@ -20,11 +21,35 @@ bool MagicItem::IsWardSpell() const noexcept
...
+    static bool runonce = true; 
+    static uint32_t bosFormID = 0;      // Can be an array if more are found.
+    if (runonce)
+    {
+        runonce = false;  
+        Mod* pBOSMod = nullptr;
+        MagicItem* pSpell  = nullptr;
+        if (   (pBOSMod = ModManager::Get()->GetByName("ccbgssse038-bowofshadows.esl")) 
+            && (pSpell  = Cast<MagicItem>(TESForm::GetById(pBOSMod->GetFormId(0x805)))))
+            bosFormID = pSpell->formID;
```
Force67, line 42: Are we aware of some more keywords which will require special cases in the future? in that case, maybe lets move it into a fn that then gets call::onced to set a static member variable for the form id(s)?

Force67 (changes requested):

Code/client/Games/Skyrim/Forms/MagicItem.cpp
rfortier: Sure. I thought this flow was single-threaded, though? 
Or maybe you just prefer call_once pattern to winging it.

Code/client/Games/Skyrim/Forms/MagicItem.cpp
rfortier: No other examples have been detected so far, and it has been months. Still, an std::call_once() pattern is going to push it that way anyway. I still wouldn't bother putting in the 1-long array and iterator at this stage.

Force67: alright, but we refactor if we hit more cases. for now i approve

PR #811, opened 2025-09-23
Reduce the amount of sliding NPCs by caching and replaying animations

miredirex:
"Sliding NPCs" refers to NPCs with a desynced animation state. You may have seen instances where they slide, skate, T-pose, stand on chairs, or move/run while seated and the like. Having a single `LatestAction` was not enough to bring Actors into the correct animation state. Now non-cell-owners (aka party members) will replay recently run animations when an Actor is being set up in their world.

Unfortunately, these changes do not fully resolve the issue - the replay is only collected while a player is connected to the server, and some animations/actions may have already run offline.

This PR also reverts #762, as with the addition of `EarlyAnimationBufferComponent` storing `LatestWeapEquipAnimation` is no longer necessary.

(continuation and rewrite of #757)

### DIFF
--- a/Code/server/Game/Animation/ActionReplayCache.cpp
+++ b/Code/server/Game/Animation/ActionReplayCache.cpp
@@ -37,10 +37,10 @@ bool ActionReplayCache::RefineReplayCache() noexcept
         ActionEvent& action = m_actions[i];
 
         // Instant counterparts of actions are highly preferred for animation replay
-        if (std::optional<String> instantAnimName = FindInstantCounterpartForAction(action.EventName))
+        if (auto instantAnimName = FindInstantCounterpartForAction(action.EventName))
         {
-            action.EventName = *instantAnimName;
-            action.TargetEventName = *instantAnimName;
+            action.EventName = String{*instantAnimName};
+            action.TargetEventName = String{*instantAnimName};
             action.ActionId = 0;
             action.IdleId = 0;
         }
@@ -63,10 +63,10 @@ bool ActionReplayCache::RefineReplayCache() noexcept
 
 bool ActionReplayCache::IsExitAction(const ActionEvent& acAction) noexcept
 {
-    if (AnimationEventLists::g_actionsExitSpecial.contains(acAction.EventName))
+    if (AnimationEventLists::kExitSpecial.contains(acAction.EventName))
         return true;
 
-    // Otherwise check string suffix
+    // Not found in special exit actions, let's check the suffix
     const String& originalName = acAction.EventName;
     String lowercaseName{acAction.EventName};
     std::transform(originalName.begin(), originalName.end(), lowercaseName.begin(),
@@ -78,19 +78,19 @@ bool ActionReplayCache::IsExitAction(const ActionEvent& acAction) noexcept
 
 bool ActionReplayCache::ShouldIgnoreAction(const ActionEvent& acAction) noexcept
 {
-    return AnimationEventLists::g_actionsIgnore.contains(acAction.EventName);
+    return AnimationEventLists::kIgnore.contains(acAction.EventName);
 }
 
-std::optional<String> ActionReplayCache::FindInstantCounterpartForAction(const String& acAction) noexcept
+std::optional<std::string_view> ActionReplayCache::FindInstantCounterpartForAction(std::string_view aAction) noexcept
 {
-    auto it = AnimationEventLists::g_actionsIdleToInstant.find(acAction);
-    if (it != AnimationEventLists::g_actionsIdleToInstant.end())
+    auto it = AnimationEventLists::kIdleToInstant.find(aAction);
+    if (it == AnimationEventLists::kIdleToInstant.end())
+        return std::nullopt;
+
+    auto& [_, instantAnimationName] = *it;
+    if (!instantAnimationName.empty())
     {
-        auto& [_, instantAnimationName] = *it;
-        if (!instantAnimationName.empty())
-        {
-            return {instantAnimationName};
-        }
+        return {instantAnimationName};
     }
     return std::nullopt;
 }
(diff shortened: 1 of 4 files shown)


Code/server/Game/Animation/ActionReplayCache.cpp
```diff
@@ -0,0 +1,96 @@
...
+bool ActionReplayCache::RefineReplayCache() noexcept
+{
+    int32_t dropAllUpToIndex = -1;
+
+    for (int32_t i = m_actions.size() - 1; i >= 0; --i)
+    {
+        ActionEvent& action = m_actions[i];
+
+        // Instant counterparts of actions are highly preferred for animation replay
+        if (std::optional<String> instantAnimName = FindInstantCounterpartForAction(action.EventName))
```
Force67: nit: an auto would be ok here too

Code/server/Game/Animation/ActionReplayCache.cpp
```diff
@@ -0,0 +1,96 @@
...
+    if (dropAllUpToIndex == -1)
+        return false;
+
+    m_actions.erase(m_actions.begin(), m_actions.begin() + dropAllUpToIndex);
+    return true;
+}
+
+bool ActionReplayCache::IsExitAction(const ActionEvent& acAction) noexcept
+{
+    if (AnimationEventLists::g_actionsExitSpecial.contains(acAction.EventName))
```
Force67: What benefit does the hardcoded set bring here? since its a const array, we dont care about the uniqueness? wouldnd an std::array suffice?

Code/server/Game/Animation/ActionReplayCache.cpp
```diff
@@ -0,0 +1,96 @@
...
+    std::transform(originalName.begin(), originalName.end(), lowercaseName.begin(),
+                   [](unsigned char c) { return std::tolower(c); });
+
+    return lowercaseName.starts_with("idle") &&
+           (lowercaseName.ends_with("exit") || lowercaseName.ends_with("exitstart"));
+}
+
+bool ActionReplayCache::ShouldIgnoreAction(const ActionEvent& acAction) noexcept
+{
+    return AnimationEventLists::g_actionsIgnore.contains(acAction.EventName);
```
Force67: same here

Code/server/Game/Animation/ActionReplayCache.cpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+std::optional<String> ActionReplayCache::FindInstantCounterpartForAction(const String& acAction) noexcept
+{
+    auto it = AnimationEventLists::g_actionsIdleToInstant.find(acAction);
```
Force67: just early return if it == end then do the rest, imho cleaner

Code/server/Game/Animation/AnimationEventLists.cpp
```diff
@@ -0,0 +1,188 @@
...
+    {"IdleLuteStart", ""},
+    {"IdleReadElderScrollStart", ""},
+    {"IdleTableDrinkAndMugEnter", ""},
+    {"IdleTableDrinkEnter", ""},
+    {"IdleTableMugEnter", ""},
+    {"IdleTablePassOutEnter", ""},
+    {"IdleWarmHands", ""},
+};
+
+const Set<String> AnimationEventLists::g_actionsExitSpecial = {
```
Force67: Lets use an std::array, we can even make that constexpr. This will generate a initterm construtor. Uniqueness should be enforced by the dev adding these strings, so a set is overkill.

Code/server/Game/Animation/AnimationEventLists.cpp
```diff
@@ -0,0 +1,188 @@
...
+    {"StoolBackExit"},
+    {"TableBackExit"},
+    {"TrapExitStart"},
+    {"VehicleExit"},
+    {"WebIdleExit"},
+    {"GetUpBegin"}, // Not an "Exit" action per se, but belongs here as it makes the actor return to its initial anim state
+    // ...others are checked in `IsExitAction()` using ends_with
+};
+
+const Set<String> AnimationEventLists::g_actionsIgnore = {
```
Force67: same here, constexpr std::array

Code/server/Game/Animation/AnimationEventLists.cpp
```diff
@@ -0,0 +1,188 @@
+#include "AnimationEventLists.h"
+
+/*
+/* The lists in this file might not contain all relevant animation events, so extend as necessary
+*/
+
+const Map<String, String> AnimationEventLists::g_actionsIdleToInstant = {
```
Force67: While a map lookup is "fast" a constexpr arr lookup may be faster: struct Pair { std::string_view from, to; };

// Keep sorted by `from` (lexicographically) and unique
inline constexpr std::array<Pair, /*N*/> kIdleToInstant = {{

Code/server/Game/Animation/AnimationEventLists.cpp
Force67: Also, avoid String, use std::string_view or a const_char* ptr, the data lives in .rdata anyway

Code/server/Game/Animation/AnimationEventLists.cpp
Force67: Also, avoid String, use std::string_view or a const_char* ptr, the data lives in .rdata anyway

Code/server/Game/Animation/AnimationEventLists.cpp
Force67: Also, avoid String, use std::string_view or a const_char* ptr, the data lives in .rdata anyway

Code/server/Game/Animation/ActionReplayCache.cpp
Force67: How often is that path hit?

Code/server/Game/Animation/AnimationEventLists.cpp
Force67: we used KSomethingSomething for konstant globals.

PR #815, opened 2025-10-14
Fix quest operator== that can rarely cause failure to progress quest

rfortier:
Self explanatory?

Found by @miredirex working on misc. quest sync.


miredirex:
> that can rarely cause failure to progress quest

Is this == of this particular struct even invoked anywhere in the code? If not, then it never caused issues. But a good fix nonetheless

miredirex (approved):

PR #816, opened 2025-10-15
Fix/engine fixes v7 support

rfortier:
SSE Engine Fixes support updated for the v7.0.19+ releases of EngineFixes. V6 is still supported (for now), but an open question is should we delete v6 support. Simpler code for us, and v7 does seem to fix more things than v6.

Summary of changes:

- Figured out how to support IAT hooks that load _during_ SkyrimSE.exe loading. All that was needed was to rearrange a few lines in ExeLoader::Load(). This not only supports skse64_plugin_loader (d3dx9_42.dll) that EngineFixes uses for early-stage patches, it now works for anything else beyond EngineFixes that uses the plugin loader or similar techniques, including CommonLibSSE variants that can be difficult to ferret out due to static linking.
- With that fixed, discovered that one of EF's memory patches smashes STR's HookFormAllocator hook; the one that allocates the ActorExtension. If it gets smashed, put it back to downcall to EF (or whatever is next in the hook list).
- And with that fixed, found that all EF patches (or, at least all turned on in the default config file) are working. Changed the "dll greylisting" code to check if the config is for EF v7, and if so, don't require any changes.

Credit @miredirex for a better hook change detector / repairer.

### DIFF
--- a/Code/client/Games/Memory.cpp
+++ b/Code/client/Games/Memory.cpp
@@ -67,33 +67,37 @@ void Memory::Free(void* apData) noexcept
     TiltedPhoques::ThisCall(RealFormFree, GameHeap::Get(), apData, false);
 }
 
-bool Memory::IsFormAllocateReplacedByEF() noexcept
+static bool IsFormAllocateReplacedByEF(TFormAllocate** appOutEngineFixesAlloc) noexcept
 {
     POINTER_SKYRIMSE(TFormAllocate, s_formAllocate, 68115);
     TFormAllocate* pFormAllocate = s_formAllocate.Get();
 
-    // 6-byte sequence of 'FF 25 00 00 00 00' comes before the virtual address we're after; so, +6
-    auto possibleEfAllocAddr = *reinterpret_cast<uintptr_t*>(reinterpret_cast<uint8_t*>(*pFormAllocate) + 6);
+    auto opcodeBytes = reinterpret_cast<uint16_t*>(*pFormAllocate);
+    uint8_t shift = 0;
+
+    if (*opcodeBytes == 0x25FF) // 'jmp' opcode 'FF 25' and the 4-byte displacement bytes come before the virtual address we're after
+        shift = 6;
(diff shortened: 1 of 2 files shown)


Code/client/TiltedOnlineApp.cpp
```diff
@@ -52,6 +52,9 @@ void* TiltedOnlineApp::GetMainAddress() const
 
 bool TiltedOnlineApp::BeginMain()
 {
+    if (GetModuleHandleW(L"EngineFixes.dll") && Memory::IsFormAllocateReplacedByEF())
```
Force67: Since this is used twice, lets make a NeedsEFPatch Free fn?

miredirex: we're working on https://github.com/rfortier/TiltedEvolution-rwf/pull/4

rfortier: Yeah, that’s actually fine now. You caught it between the time I thought I had merged M’s latest and the time I fixed the merge. 

There is only one appearance, in memory.cpp

Code/client/Games/Memory.cpp
```diff
@@ -67,6 +67,39 @@ void Memory::Free(void* apData) noexcept
...
+    uint8_t shift = 0;
+
+    if (*opcodeBytes == 0x25FF) // 'jmp' opcode 'FF 25' and the 4-byte displacement bytes come before the virtual address we're after
+        shift = 6;
+    else if (*opcodeBytes == 0xB848) // 'mov' opcode '48 B8' comes before the virtual address we're after
+        shift = 2;
+
+    auto possibleEfAllocAddress = *reinterpret_cast<uintptr_t*>(reinterpret_cast<uint8_t*>(*pFormAllocate) + shift);
+
+    MEMORY_BASIC_INFORMATION mbi;
```
Force67: Lets zero init this for safety

Force67 (approved):

PR #822, opened 2025-11-03
Fix potential crash originating from HookSpellCast

miredirex:
This is how the game does its checks:

image

thx @absol89 for the crash log


rfortier:
Macro comment: Isn't that RE what STR is calling RealSpellCast? Or is that a snippet of another routine?
The problem is that this routine is returning a value, and STR's hook is returning void. Looks like another bug.

rfortier:
Other than the potential return value botch in the existing code, I agree your injected fix is doing the same testing as the RE snippet.

PR #828, opened 2025-11-13
Fix logic that determines whether to spoof GetModuleFileNameW(nullptr)

rfortier:
Update: TiltedCore v0.2.8 release has now been cut, so updated with that version.
PR, although complete barring more comments, will remain draft until prerequisites met:

1. xmake_repo accepts the PR for the new version of TiltedCore.
2. PRs to update other modules can then be submitted. TiltedHooks, TiltedReverse, and TiltedUI need a version pin update.
3. Those PRs merged. 
4. Then, finally, this PR can be moved converted from Draft to Submitted and merged.
---
It was depending on figuring out where it was called from, but the logic was unreliable and there is no apparent way to fix it. Multiple attempts using more sophisticated FOSS stack crawlers all failed.

Fixes: now, it switches spoofing on just before loading SkyrimSE.exe (so the loaded DLLs see the path they expect).

The required companion fix in TIltedCore:TiltedPhoques::GetPath() persists the value found originally, so it doesn't change out from under with late callers. If your code has only late callers, just add a call before loading your game binary.

### DIFF
--- a/Code/immersive_launcher/Launcher.cpp
+++ b/Code/immersive_launcher/Launcher.cpp
@@ -42,6 +42,11 @@ LaunchContext* GetLaunchContext()
     return g_context;
 }
 
+bool LaunchContext::GetLoaded()
+{
+    return isLoaded;
+}
+
 // Everything is nothing, life is worth living, just look to the stars
 #define DIE_NOW(err)  \
     {                 \
@@ -126,6 +131,7 @@ bool LoadProgram(LaunchContext& LC)
     LC.Version = QueryFileVersion(LC.exePath.c_str());
     if (LC.Version.empty())
         DIE_NOW(L"Failed to query game version");
+    LC.SetLoaded();
 
     ExeLoader loader(CurrentTarget.exeLoadSz);
     if (!loader.Load(reinterpret_cast<uint8_t*>(content.data())))
--- a/Code/immersive_launcher/Launcher.h
+++ b/Code/immersive_launcher/Launcher.h
@@ -22,6 +22,12 @@ struct LaunchContext
     fs::path gamePath;
     TiltedPhoques::String Version;
     ExeLoader::TEntryPoint gameMain = nullptr;
+
+    void SetLoaded() { isLoaded = true; }   // If loaded, need to spoof GetModuleFileName*(nullptr)
+    bool GetLoaded();
+
+  private:
+    bool isLoaded{false};
 };
 
 LaunchContext* GetLaunchContext();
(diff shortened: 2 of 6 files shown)


Code/immersive_launcher/Launcher.h
```diff
@@ -21,7 +21,11 @@ struct LaunchContext
     fs::path exePath;
     fs::path gamePath;
     TiltedPhoques::String Version;
+    bool isLoaded{false};
     ExeLoader::TEntryPoint gameMain = nullptr;
+
+    void SetLoaded() { isLoaded = true; }   // If loaded, need to spoof GetModuleFileName*(nullptr)
+    bool GetLoaded();
```
miredirex, line 27: Unnecessary getter and setter. The field is public

Code/immersive_launcher/Launcher.cpp
```diff
@@ -126,6 +131,7 @@ bool LoadProgram(LaunchContext& LC)
     LC.Version = QueryFileVersion(LC.exePath.c_str());
     if (LC.Version.empty())
         DIE_NOW(L"Failed to query game version");
+    LC.SetLoaded();
```
miredirex, line 134: Why not after the `if (!loader.Load(reinterpret_cast<uint8_t*>(content.data())))` call?

Code/immersive_launcher/Launcher.h
rfortier: fixed

Code/immersive_launcher/Launcher.cpp
rfortier: Slightly subtle. The goal is for the custom loader to load SkyrimSE.exe with spoofing enabled so any DLLs, in particular any preload DLLs like SSE EngineFixes.dll, experience what looks like SkyrimSE.exe is loading rather than experiencing SkyrimTogether.exe as loaded. 

In particular, attempts to evaluate the directory of the running executable have to return the GamePath, not the  SkyrimTogether.exe path.

So the switch must be thrown before loading the executable.

rfortier:
This PR is now ready to merge. 

Updated all submodules to latest commit, which removes the version pin on TiltedCore. This means submodules will consistently use what the parent TiltedEvolution uses, which is 0.2.8. If the submodules load an older version it conflicts and will crash.

Rebased to latest /dev to ensure no merge conflicts.

Squashed some commits.

Left the 4 submodule updates as a separate commit for visibility.

rfortier:
Sigh. Back to Draft. 

The latest TiltedUI, before the pin change, is incompatible with the latest TiltedEvolution; TiltedUI expects the keybinds PR to be merged, which it is not. So, F2 stopped working.

rfortier:
Review comment bugfix. Moving to _latest_ TiltedHooks got the version created for key rebinds, which no longer initializes the UI toggle keys. Have to initialize them if you don't have PR #724  changes

miredirex (approved):

PR #832, opened 2025-11-30
CI: build/publish Docker linux server

llerafel:


.github/workflows/docker-unstable.yml
```diff
@@ -0,0 +1,49 @@
+name: Docker build (unstable)
```
miredirex: I wouldn't name it "unstable" because we will also use it for actual big releases. Rename the yml file too, please

.github/workflows/docker-unstable.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Get version
+        id: output-version
+        run: echo "version=$(git describe --tags)" >> $GITHUB_OUTPUT
+
+      - name: Build Docker image
+        run: |
+          docker build \
+            -t ${{ secrets.DOCKERHUB_USERNAME }}/st-reborn-server:${{ steps.output-version.outputs.version }} \
+            -t ${{ secrets.DOCKERHUB_USERNAME }}/st-reborn-server:unstable-latest .
```
miredirex: We need to build and push both x86-64 and arm64 under one tag, see here - https://github.com/tiltedphoques/TiltedEvolution/blob/dev/Dockerfile#L4-L6

.github/workflows/docker-unstable.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Get version
+        id: output-version
+        run: echo "version=$(git describe --tags)" >> $GITHUB_OUTPUT
```
miredirex:
```yml
- name: Store version in env
  run: echo "VERSION=$(git describe --tags)" >> $env:GITHUB_ENV
```
Then just `${{ env.VERSION }}`

llerafel:
Thank you very much for the review.

I have updated the workflow based on your feedback:
- Renamed the workflow and file (removed "unstable") — it now supports both releases and weekly builds;
- Version is stored via $GITHUB_ENV instead of step outputs;
- Added QEMU + Buildx for multi‑arch builds;
- Split the logic into Release (tags + latest) and Weekly (weekly-latest).

Current logic:
* Release​‍​‌‍​‍‌​‍​‌‍​‍‌ images are created and uploaded to the server when a **new tag** is generated. 
* Weekly images are created **every Friday at 13:00 UTC**  (`0 13 * * 5`) - a.k.a. `Playable Skyrim Together ​‍​‌‍​‍‌​‍​‌‍​‍‌Build`.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      # (Release) Build the server for both x86_64 and arm64 platforms
+      - name: Build and push release image
+        if: startsWith(github.ref, 'refs/tags/v')
```
miredirex: `if: github.ref_type == 'tag'`

.github/workflows/docker.yml
```diff
@@ -0,0 +1,77 @@
...
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: |
+            ${{ secrets.DOCKERHUB_USERNAME }}/st-reborn-server:${{ github.ref_name }}
+            ${{ secrets.DOCKERHUB_USERNAME }}/st-reborn-server:latest
+          labels: |
+            org.opencontainers.image.title=st-reborn-server
+            org.opencontainers.image.source=${{ github.repository }}
+            org.opencontainers.image.revision=${{ github.sha }}
+            org.opencontainers.image.version=${{ github.ref_name  }}
```
miredirex:
```suggestion
            org.opencontainers.image.title=Skyrim Together Server
            org.opencontainers.image.source=${{ github.repository }}
            org.opencontainers.image.revision=${{ github.sha }}
            org.opencontainers.image.version=${{ env.VERSION  }}
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,77 @@
...
+    - cron: '0 13 * * 5'
+  push:
+    branches:
+      - master
+    tags:
+      - 'v[0-9]+\.[0-9]+\.[0-9]+'
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
```
miredirex, line 16: Image

Hmmmm... Not asking you to change this right now, but this is something to consider for faster CI runs as QEMU arm64 compilation causes tons of overhead and takes hours. This should be possible if we manage to reuse the `buildx` cache

miredirex (approved):

.github/workflows/docker.yml
Force67: Lets just run a x64 run via the regular runner, and build the arm one on the arm image directly: https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

Id like to do it properly now rather than later

llerafel: @Force67 
Thanks for the suggestion!
The plan is to handle the x64/arm split and caching/matrix optimizations in a dedicated follow‑up PR, where we can properly discuss the best way to speed things up.

For now, before Friday, the priority is to merge this PR so the scheduled builds are running.
That way we’ll have real data from the cron jobs to compare against when we start the optimization work

PR #839, opened 2025-12-09
Fix multiple starts of quests when in-party

rfortier:
An update to PR #805 that fixes some additional cases.

Raise logging level and fix logs of quest progression to enable tracking this down.

Previously, multi-execution of quest stage entry was suppressed. But multi quest starts weren't. This is a problem when a quest start has side effects that advance quest stage; as the party members started their quests, the stage-rewind would reflect back to the leader, allowing quest stage side effects to happen again.

Also added fix to debugger to force setStage. This is needed so leader can set the stage to the stage they are at, and not have it suppressed so it is sent to party.

Regularized quest logging so you can import into a spreadsheet, break fields on colon and comma (:,)

### DIFF
--- a/Code/client/Games/Skyrim/Forms/TESQuest.cpp
+++ b/Code/client/Games/Skyrim/Forms/TESQuest.cpp
@@ -97,7 +97,7 @@ bool TESQuest::SetStage(uint16_t newStage)
     return SetStage(this, newStage);
 }
 
-void TESQuest::ScriptSetStage(uint16_t stageIndex, boolean bForce)
+void TESQuest::ScriptSetStage(uint16_t stageIndex, bool bForce)
 {
     if ((currentStage == stageIndex || IsStageDone(stageIndex)) && !bForce)
         return;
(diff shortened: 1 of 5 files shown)


Code/client/Games/Skyrim/Forms/TESQuest.cpp
```diff
@@ -97,9 +97,9 @@ bool TESQuest::SetStage(uint16_t newStage)
     return SetStage(this, newStage);
 }
 
-void TESQuest::ScriptSetStage(uint16_t stageIndex)
+void TESQuest::ScriptSetStage(uint16_t stageIndex, boolean bForce)
 {
-    if (currentStage == stageIndex || IsStageDone(stageIndex))
+    if ((currentStage == stageIndex || IsStageDone(stageIndex)) && !bForce)
```
miredirex: (note to self: test cutscene without the `currentStage == stageIndex` change or whatever the blame points to)

rfortier:
Closing, this is about to be replaced with a far more complete PR.

rfortier:
Reopening. This is the most recent version of quest stage multiple-trigger suppression (such as for The Way Of The Voice) that predates the full deduplication logic that follows it (but I was optimistic about how close to completion it was). 

This one is safe to merge while still polishing its successor.

PR #847, opened 2026-01-04
Update to PR #727 better better multithread logging

rfortier:
Follow-on to PR #727

In addition to console window, make sure the file logs also contain the millisecond and thread ID.

Flushes server console log every 2 seconds.

Flushes client log on every log line at [info] or greater. This is to make sure that crash logs are up-to-date (or reasonably close).

### DIFF
--- a/Code/client/TiltedOnlineApp.cpp
+++ b/Code/client/TiltedOnlineApp.cpp
@@ -36,8 +36,6 @@ TiltedOnlineApp::TiltedOnlineApp()
     auto rotatingLogger = std::make_shared<spdlog::sinks::rotating_file_sink_mt>(logPath / "tp_client.log", 1048576 * 5, 3);
     // rotatingLogger->set_level(spdlog::level::debug);
     auto console = std::make_shared<spdlog::sinks::stdout_color_sink_mt>();
-    console->set_pattern("%^[%Y-%m-%d %H:%M:%S.%e] [%l] [tid %t] %$ %v");
-
     auto logger = std::make_shared<spdlog::logger>("", spdlog::sinks_init_list{console, rotatingLogger});
     logger->set_pattern("%^[%Y-%m-%d %H:%M:%S.%e] [%l] [tid %t] %$ %v");
     spdlog::flush_every(std::chrono::seconds(1));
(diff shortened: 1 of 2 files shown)


miredirex (commented):
> Flushes server console log every 2 seconds.

Could it cause lag? Or does it happen in background?

> This is to make sure that crash logs are up-to-date (or reasonably close).

https://github.com/tiltedphoques/TiltedEvolution/blob/cecc9c8430234e5ebe95e8b55955f4ee5edf9417/Code/client/CrashHandler.cpp#L100-L104

I think we already flush logs on any crash

rfortier:
> > Flushes server console log every 2 seconds.
> 
> Could it cause lag? Or does it happen in background?

It won't cause lag. Doesn't for the console window, which flushes every line. For flush_every(), flushing is done on an async thread. 

Without this the console logfile is only flushed when it fills up a page, which is 8 or 16KB (so, could be very out-of-date). And, if you finish a session and kill the console window, it is never flushed.

> https://github.com/tiltedphoques/TiltedEvolution/blob/cecc9c8430234e5ebe95e8b55955f4ee5edf9417/Code/client/CrashHandler.cpp#L100-L104
> 
> I think we already flush logs on any crash

Well, yeah. I'm pretty sure I put that line there when I made crashloggers work. 
It doesn't work, though. Probably because it is flushing the wrong logger, I didn't know as much at the time. Or, because we're crashing and the logger is borked. Keeping the log reasonably up-to-date should work around the issue regardless.

But I will take a look and see which stream is the file logger; fixing that flush would be nice. The tweak as-is is still needed to flush the file, though, or things that are tailing/watching the log file won't be up to date.

rfortier:
It's been stable. Added one last tweak (date, helps us reject people sending wrong logs)

miredirex:
Suggestion: `"%^[%Y-%m-%d %H:%M:%S.%e]` for the log files, but not for the console output. This should be possible with spdlog

rfortier:
Updated to flush_on( 1 second ) instead. Good enough for  an active tail -f on the log file, and infinitesimal overhead. 

Will need to separately figure out why the flush in the crash path is not working. The correct sink is being flushed, but the log file is not being updated.

PR #849, opened 2026-01-12
Improve scripted cutscenes sync where player controls are disabled

miredirex:
There are certain quests where the game forcefully disables controls for the player(s). In that case, only the party leader should be the one advancing quests forward (i.e. broadcasting quest stage updates to others)

This PR addresses a bug introduced by #769


miredirex:
Closing in favor of #848

PR #852, opened 2026-01-25
Basic .clang-format fixes

rfortier:
Drop .clang-format ColumnLimit to 180 (from 300) so lines fit on less than an ultra-wide.
Delete incorrect .clang-format accidently committed years ago.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -12,7 +12,7 @@ BraceWrapping:
   IndentBraces: true
 BreakBeforeBraces: Allman
 BreakConstructorInitializers: BeforeComma
-ColumnLimit: 300
+ColumnLimit: 180
 IndentWidth: 4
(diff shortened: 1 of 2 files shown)


Force67:
Alongside with this we should ensure that our entire ocde is properly formatted, and that CI rejects badly formatted PRs imho. What does @miredirex think abt that?

miredirex:
Yeah I will get to that, but after we release 1.8.1

rfortier:
Yes, we've been talking about that. This is just a warmup, the simplest/safest possible fix so clang-format works in Visual Studio again (without manual fixups to the tree that you tend to lose on branch changes). Then @miredirex was going to tweak the settings for a bigger fix. But again, after the next dot release; baby steps.

PR #855, opened 2026-02-05
Patch Companions C02 "Brotherhood" (induction) so member does not hang

rfortier:
Patches the single Scene in the quest to fix the hang if a member (not Leader) steps on the ceremony hotspot. The fix is innocuous to the standard game.

It was an obscure bug. There's a hidden (not in the the quest log, a bit of an easter egg) objective to speak to the QuestGiver (Vilkas) before going to the ceremony.

IF the Leader does not talk to Vilkas
AND the member steps on the ceremony hotspot
AND the Leader does not step on the hotspot

Scene would hang, so the quest would hang with player controls frozen. The reason is the scene required Vilkas to be present. The hidden objective makes him be present. If the Leader triggered the ceremony (or joins it), Vilkas is told to run to the ceremony.

If Member triggered it, well, Vilkas did nothing because he's controlled by Leader.

Isn't this fun? Anyway, fixed it by only requiring always-present NPCs to speak their last line.

Separate topic: I made this as a separate ESL-flagged ESP. I think now that ESLs exist and are cheap, not burning slots, it makes more sense to keep patches separate rather than bundling them all into one ESP glob and losing context of what is changing what.

But that's debatable, so if there are strong arguments to merge every patch into one .ESP glob, let's hear them.


miredirex:
I vote for having a single central `SkyrimTogetherQuestPatches.esp`. Don't forget to also update this list:

https://github.com/tiltedphoques/TiltedEvolution/blob/97e9d281e25ab266d1de0c34cdaee7ebd6a4b286/Code/client/Services/Generic/DiscoveryService.cpp#L257-L259

**upd.** [Spriggit](https://github.com/Mutagen-Modding/Spriggit) is something to consider (in the future)

PR #858, opened 2026-02-23
IsSlowTime fix caster, not target

rfortier:
A fix to a prior fix, which observed SlowEffect was doubled if cast by an NPC. 
Bug was changing TargetId instead of the CasterId. It does the latter now.

### DIFF
--- a/Code/client/Services/Generic/MagicService.cpp
+++ b/Code/client/Services/Generic/MagicService.cpp
@@ -408,6 +408,15 @@ void MagicService::OnNotifyAddTarget(const NotifyAddTarget& acMessage) noexcept
         return;
     }
 
+    // This hack is here because slow time seems to be twice as slow when cast by an npc
+    if (pEffect->IsSlowEffect())
+    {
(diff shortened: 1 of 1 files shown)


miredirex:
https://github.com/tiltedphoques/TiltedEvolution/commit/a3d9f5dd89d719bc4e8ee08a0177d6b83c8de6d4 "fix: slow time shout sync"

Did you test [the shout](https://en.uesp.net/wiki/Skyrim:Slow_Time)?

rfortier:
> [a3d9f5d](https://github.com/tiltedphoques/TiltedEvolution/commit/a3d9f5dd89d719bc4e8ee08a0177d6b83c8de6d4) "fix: slow time shout sync"
> 
> Did you test [the shout](https://en.uesp.net/wiki/Skyrim:Slow_Time)?

nope. I assume slow time shout specifically. I can take a look at it the next time I have any time for STR hacking.

Do we think slow time will even sync? it's probably a magic effect but not all of them work.

miredirex:
See #886, closing for now

PR #862, opened 2026-03-24
fix: null deref in OnRequestRespawn

YuneGit:
try_get<AnimationComponent>() on line 502 of CharacterService.cpp was being dereferenced without a null check. The view only includes OwnerComponent and CharacterComponent, so AnimationComponent isn't guaranteed to exist on the entity. Added a null check to match every other try_get usage in the file.


YuneGit:
bad formatting, ignore this

miredirex:
Did you actually have this server crash?

Issue #845, opened 2025-12-31
Prevent anyone besides party leader from initially entering zone transition doors


uuRETCH:
Unsure of how feasible this would be, but I feel it would prevent a lot of accidents if party members were altogether denied from using any zone transition door that was not used by the party leader.

Zone doors used by the party leader could be highlighted once used, in similar fashion to the 'Reveal Players' function. Signals to party members where to go if they've somehow gotten too far separated in the area.

miredirex:
Not a bad idea, but perhaps too restrictive. In my opinion, best course of action is to first resolve the bugs surrounding NPC ownership transfer so that cell changes become less fragile and desync-prone

Relevant cosi comment (this has been discussed in `#questions` a while ago):

<img width="774" height="161" alt="Image" src="https://github.com/user-attachments/assets/39d8eebf-eaa8-439f-a5e3-a854809afcba" />

ExAdventurer:
I just came up with this solution independently after testing STR with a friend. Found this after posting in `#questions` myself. 

I second this solution.

PR #865, opened 2026-04-20
Prevent non-leader party members from using zone transition doors (#845)

sfedev:
When the local player is in a party but is not the leader, block load door activations (Door + ExtraTeleport) and show a system message pointing the user at the party menu teleport. Leader activations, solo play, non-party contexts, NPC/script activations, and in-cell doors (which lack ExtraTeleport) are untouched.
Scope note: this is the minimal "always block" interpretation. The issue's richer "door unlocks for followers once the leader passes through" mechanic is deferred, since it needs per-door state synced across the party and is worth its own PR.
Known trade-off: a quest script that calls `player.Activate(loadDoor)` directly will also be blocked for non-leaders. This path is rare in practice; if it surfaces, a per-quest exclusion list (mirroring QuestService's kNonSyncableQuestIds) is the natural follow-up.

### DIFF
--- a/Code/client/Games/Skyrim/TESObjectREFR.cpp
+++ b/Code/client/Games/Skyrim/TESObjectREFR.cpp
@@ -983,7 +983,7 @@ bool TP_MAKE_THISCALL(HookActivate, TESObjectREFR, TESObjectREFR* apActivator, u
     Actor* pActivator = Cast<Actor>(apActivator);
 
     // Exclude books from activation since only reading them removes them from the cell
-    // Note: Books are now unsynced 
+    // Note: Books are now unsynced
     if (pActivator && apThis->baseForm->formType != FormType::Book)
     {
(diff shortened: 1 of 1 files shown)


miredirex:
What about the caves and other means of entering locations?

Also
> `"Only the party leader can use zone transition doors. Use the party menu to teleport to them."`

needs to be a [localizable string](https://github.com/tiltedphoques/TiltedEvolution/blob/dev/Code/skyrim_ui/src/assets/i18n/en.json).

As I already commented on #845, we should really just prioritize fixing robustness of the existing actor management system. This change makes the game very restrictive (although maybe for good), I'd like to hear more opinions on this.

Corbent:
Opinion: This could be good outside of towns. But in cities it feels good to be able to disperse and each person carry on their business with vendors and such. Then regroup o continue the adventure.

PR #868, opened 2026-05-27
Hide location from player list; default to party menu tab to prevent trolling

absol89:
On servers with auto-party join turned off, this will help prevent location trolling on public servers, they join and kill npcs where the existing players are located. This used to be displayed in the player list and even if denied party, trolls could go there and kill.

Also, the party menu is infinitely more useful as a default tab in the player manager. So I thought to make it the default tab too.


absol89:
Tested 16 june after successful playable windows build artifact was added to vortex with skse and address library 1.6

Party leader view, opened party tab by default, started a party, when other player joined, location is visible to party.
image

View of the censored player list
image

All party interactions like kick and teleport were verified in the test @miredirex the builds work except for Engine fixes.

PR #869, opened 2026-05-27
Add per-party auto-join toggle (AllowAutoJoin) with UI checkbox

absol89:
Problem: The server-wide bAutoPartyJoin setting cannot be overridden per-party. When enabled, all new players auto-join the sole party on the server. Party leaders cannot opt out without disabling the setting globally. This lets trolls repeatedly grief party cells on public servers, such as joining with enchantment mods and teleporting to players who crash, or killing their owned NPCs.

Solution: Add a per-party AllowAutoJoin flag (default true, matching bAutoPartyJoin). The party leader can toggle it at runtime via an 'Auto Join' checkbox in the Party Menu UI. When disabled, the party is excluded from auto-join for newly connecting players.

Changes:
- Protocol: New PartyAutoJoinToggleRequest message + opcode. NotifyPartyInfo extended with AllowAutoJoin, ServerAutoJoin, PartyCount.
- Server: Party struct gains AllowAutoJoin{true}. OnPartyCreate sets AllowAutoJoin = bAutoPartyJoin. OnPartyAutoJoinToggle handler lets the leader flip it. OnPlayerJoin skips auto-join if AllowAutoJoin is false. OnRemovePlayerFromParty broadcasts remaining party info when a party is destroyed and only 1 remains.
- C++ client: Tracks AllowAutoJoin, ServerAutoJoin, PartyCount from NotifyPartyInfo, forwards all to JS overlay.
- Angular UI: PartyInfo and Group models carry allowAutoJoin, serverAutoJoin, partyCount. Party Menu shows an 'Auto Join' checkbox (leader-only) that is:
  * Checked when AllowAutoJoin is true, this is the default value for the first party on enabled servers.
  * Disabled when ServerAutoJoin is false OR PartyCount > 1 Translated in all 13 supported languages. 
  * The checkbox remains unchecked on unlock, when the parties drop from 2 to 1, but can be interacted with to toggle again.
  * If all parties are disbanded, and a new one is created, it will adhere to the serverAutoJoin bool again.


absol89:
Comments about the full lifecycle:

    Before connecting / at app start:
    - Group defaults: allowAutoJoin=false, serverAutoJoin=false, partyCount=0
    - Checkbox doesn't show because isPartyLeader$ is false (not in a party)

    After connecting, before creating a party:
    - Still no group state — checkbox not visible

    After creating or joining a party (BroadcastPartyInfo fires):
    - allowAutoJoin = server's bAutoPartyJoin value (true by default)
    - serverAutoJoin = server's bAutoPartyJoin value
    - partyCount = current number of parties on server
    - If you're the leader: checkbox appears, checked if allowAutoJoin && serverAutoJoin && partyCount==1

    Leader toggles checkbox off → ToggleAutoJoinParty():
    - Server flips AllowAutoJoin to false
    - BroadcastPartyInfo fires → clients update
    - Checkbox shows unchecked

    New player joins server with 1 party, bAutoPartyJoin=true:
    - OnPlayerJoin checks AllowAutoJoin — if false, skips auto-join
    - Player is NOT auto-joined into the party

    2nd party created on server:
    - Both parties' BroadcastPartyInfo fires with partyCount=2
    - Checkboxes on both parties get disabled (partyCount > 1)

    2nd party destroyed → only 1 remains:
    - RemovePlayerFromParty erases the empty party
    - Detects m_parties.size()==1, calls BroadcastPartyInfo on remaining party
    - Remaining party's clients see partyCount=1 → checkbox becomes enabled again
    - AllowAutoJoin on the remaining party was preserved throughout

    All parties disbanded → new party created:
    - New party gets AllowAutoJoin = bAutoPartyJoin (fresh from server setting)
    - Old AllowAutoJoin state is completely gone, so future play sessions don't need cleanups, restarts or workarounds.

    The false defaults are supposedly harmless because the checkbox is gated on isPartyLeader$ which is only true when you're actually in a party as leader — and at that point the server has already informed the client of the real values. The checkbox is also not shown to party members, unless they leave and create a new party, in which case the box updates.

absol89:
I will do a simpler approach that doesnt merge conflict with other initiatives (too many files touched here)

PR #880, opened 2026-07-11
Fix/improved naked npc fixes for NPC and Players

absol89:
Long story short, rfortier discovered that the original RunNakedNPCBugChecks() didn't solve the issue. Checked every 3s.

What I discovered when testing the PR code, was that there were still cases where NPC were naked during ownership flip.

On some players, this ownership flip left local characters in a RemoveAllItems state, as inventory or equip calls sent empty.

What my commits since then have been trying to do is the following:
1. Make sure empty garbage equip/inventory states are not allowed to propagate, and not allowed to RemoveAllItems on ai.
2. Make sure local NPC actors do not send too much networking data when there's fighting for ownership. Tested in modlist.
3. Clean out old diagnostics and remove the RunNakedNPCBugChecks() code. Newer code runs on spawn + ownership flips.
4. Previously, there were reports that pickpocketing equipped garments from NPC would reset them back to a normal outfit.
5. I checked this after making all these commits, and verified that new code lets Party leader stealing worn items sync online.
6. For the final issue, players who wore no body armor or body garment, had invisible equipped hands and feet on remotes.
7. To fix this, i looked up a garment that shows an undressed torso, but counts as a body torso armor, so all gear show again.

When comparing before and after https://github.com/absol89/TiltedEvolutionScriptFixes/commit/ed7ed1ec3af5c4c9d0a42d0d1e47349f611732d3 , the previous commit's code isRemote() revealed NPC with bad worn state. That gave them the vampire lord's armor as proof that this was what was going on. Now, I changed the final commit to only give isRemotePlayer() the vampire lord armor to make sure their other limbs and helmets would render when they travel shirtless. Hopefully this fixes 3 issues in one swoop, and I am submitting it as a separate PR than 860 so there's possibility to review separately.

Video of npc not being naked after fixes. You can watch it without sound, it is just a 11 minute stress test https://www.youtube.com/watch?v=mCgSCQRPPiQ 

Video of pickpocket sync working (leader did it). Watch with sound https://www.youtube.com/watch?v=tFoMUSwnAZo

Picture of an NPC being fully equipless by player pickpocket. It was important that this didn't trigger outfit reset, seems ok.
image

Picture of how player's look when not wearing torso item, and wearing helmet, gloves and boots on vanilla STR 1.8.0 (live).
image

Video of fix to limb rendering when player is not wearing torso items (gets armor 02011a85). https://youtu.be/EfgXv9WK8fE

### DIFF
--- a/Code/client/Games/ActorExtension.h
+++ b/Code/client/Games/ActorExtension.h
@@ -17,9 +17,11 @@ struct ActorExtension
     bool IsLocalPlayer() const noexcept;
     void SetRemote(bool aSet) noexcept;
     void SetPlayer(bool aSet) noexcept;
+    void SetNakedDeadline() noexcept { nakedDeadline = std::chrono::steady_clock::now();  }
 
     ActionEvent LatestAnimation{};
     size_t GraphDescriptorHash = 0;
+    std::chrono::steady_clock::time_point nakedDeadline{};
 
   private:
     uint32_t onlineFlags{0};
--- a/Code/client/Games/ActorExtension.h
+++ b/Code/client/Games/ActorExtension.h
@@ -17,11 +17,9 @@ struct ActorExtension
     bool IsLocalPlayer() const noexcept;
     void SetRemote(bool aSet) noexcept;
     void SetPlayer(bool aSet) noexcept;
-    void SetNakedDeadline() noexcept { nakedDeadline = std::chrono::steady_clock::now();  }
 
     ActionEvent LatestAnimation{};
     size_t GraphDescriptorHash = 0;
-    std::chrono::steady_clock::time_point nakedDeadline{};
 
   private:
     uint32_t onlineFlags{0};
--- a/Code/client/Games/PapyrusFunctions.cpp
+++ b/Code/client/Games/PapyrusFunctions.cpp
@@ -7,8 +7,6 @@ namespace PapyrusFunctions
 
 bool IsRemotePlayer(Actor* apActor)
 {
-    spdlog::info("Calling IsRemotePlayer");
-
     auto* pExtension = apActor->GetExtension();
     if (!pExtension)
         return false;
--- a/Code/client/Games/Skyrim/Actor.cpp
+++ b/Code/client/Games/Skyrim/Actor.cpp
@@ -681,10 +681,20 @@ Inventory Actor::DeriveOutfitInventory() const noexcept
             // Collect the armor(s) this outfit item resolves to. A leveled list entry
             // must resolve to EXACTLY ONE pick, not every candidate: pushing all list
             // members gave an NPC one cuirass per possible roll (issue #5's stacked
-            // outfit items). Taking the first valid member is deterministic -- leveled
-            // list order is identical in every client's load order, so both clients
-            // derive the same item and snapshots stay consistent across ownership flips.
+            // outfit items). Prefer the first MAPPED BODY-PIECE member (fall back to
+            // first mapped member): leveled-list order is identical across clients, so
+            // the pick stays deterministic while guaranteeing outfits resolve to a
+            // torso -- a boot-only pick left city NPCs permanently half-naked because
+            // every downstream check tests the body slot.
             Vector<TESObjectARMO*> armors;
+
+            auto& modSystem = World::Get().GetModSystem();
+            auto isMapped = [&modSystem](TESObjectARMO* apArmor)
+            {
+                GameId id{};
+                modSystem.GetServerModId(apArmor->formID, id);
+                return id != GameId{};
+            };
 
             if (pItem->formType == FormType::Armor)
             {
@@ -706,27 +716,36 @@ Inventory Actor::DeriveOutfitInventory() const noexcept
                             if (pList[i].pForm)
                             {
                                 if (TESObjectARMO* pArmor = Cast<TESObjectARMO>(pList[i].pForm))
-                                {
                                     armors.push_back(pArmor);
-                                    break; // single deterministic roll per list
-                                }
                             }
                         }
                     }
                 }
             }
 
-            for (TESObjectARMO* pArmor : armors)
+            TESObjectARMO* pPick = nullptr;
+            for (TESObjectARMO* pCandidate : armors)
+            {
+                // Unmapped armor (mod not registered with the server) resolves to a
+                // {0,0} BaseId which collides with every other unmapped entry under
+                // dedup -- committing one erases distinct pieces. Skip them.
+                if (!isMapped(pCandidate))
+                    continue;
+                pPick = pCandidate;
+                if (pPick->IsBodyPiece())
+                    break;
+            }
+
+            if (pPick)
             {
                 Inventory::Entry entry{};
-                World::Get().GetModSystem().GetServerModId(pArmor->formID, entry.BaseId);
+                World::Get().GetModSystem().GetServerModId(pPick->formID, entry.BaseId);
                 entry.Count = 1;
                 entry.ExtraWorn = true;
                 fallback.Entries.push_back(std::move(entry));
 
                 ++wearable;
-                if (entry.BaseId != GameId{})
-                    ++mapped;
+                ++mapped;
             }
         }
 
@@ -734,8 +753,24 @@ Inventory Actor::DeriveOutfitInventory() const noexcept
                      pOutfit->formID, pOutfit->outfitItems.length, wearable, mapped,
                      fallback.HasWornItems());
 
-        // Use the first outfit slot that yields worn items; otherwise keep trying.
-        if (fallback.HasWornItems())
+        // Accept this outfit slot only if it yields a wearable set INCLUDING a body
+        // piece; otherwise try the next slot/template. A non-body-only derive (boot,
+        // satchel) is what stranded the city NPCs in the unrecoverable state.
+        bool hasBody = false;
+        for (const auto& entry : fallback.Entries)
+        {
+            const uint32_t gameId = World::Get().GetModSystem().GetGameId(entry.BaseId);
+            if (TESObjectARMO* pArmor = Cast<TESObjectARMO>(TESForm::GetById(gameId)))
+            {
+                if (pArmor->IsBodyPiece())
+                {
+                    hasBody = true;
+                    break;
+                }
+            }
+        }
+
+        if (fallback.HasWornItems() && hasBody)
             break;
     }
 
--- a/Code/client/Games/Skyrim/Actor.cpp
+++ b/Code/client/Games/Skyrim/Actor.cpp
@@ -678,8 +678,12 @@ Inventory Actor::DeriveOutfitInventory() const noexcept
             if (!pItem)
                 continue;
 
-            // Collect every armor this outfit item can resolve to: a direct ARMO, or
-            // any entry in a leveled item list (Skyrim only uses list A).
+            // Collect the armor(s) this outfit item resolves to. A leveled list entry
+            // must resolve to EXACTLY ONE pick, not every candidate: pushing all list
+            // members gave an NPC one cuirass per possible roll (issue #5's stacked
+            // outfit items). Taking the first valid member is deterministic -- leveled
+            // list order is identical in every client's load order, so both clients
+            // derive the same item and snapshots stay consistent across ownership flips.
             Vector<TESObjectARMO*> armors;
 
             if (pItem->formType == FormType::Armor)
@@ -702,7 +706,10 @@ Inventory Actor::DeriveOutfitInventory() const noexcept
                             if (pList[i].pForm)
                             {
                                 if (TESObjectARMO* pArmor = Cast<TESObjectARMO>(pList[i].pForm))
+                                {
                                     armors.push_back(pArmor);
+                                    break; // single deterministic roll per list
+                                }
                             }
                         }
                     }
--- a/Code/client/Games/Skyrim/Actor.cpp
+++ b/Code/client/Games/Skyrim/Actor.cpp
@@ -625,7 +625,7 @@ Inventory Actor::GetActorInventory() const noexcept
     // unaffected.
     if (!GetExtension()->IsPlayer() && !IsDead() && !inventory.HasWornItems())
     {
-        spdlog::info("[NakedFix] fallback path for actor {:X} (base {:X})", formID,
+        spdlog::debug("[NakedFix] fallback path for actor {:X} (base {:X})", formID,
                      baseForm ? baseForm->formID : 0);
 
         Inventory fallback = DeriveOutfitInventory();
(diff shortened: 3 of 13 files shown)


absol89:
## One-line summary for the merge

This branch fixes the root cause of locally-owned NPCs rendering naked (server applies a transient naked snapshot during ownership/cell-reload; engine strips dressed NPCs on 3D-load) by guarding those two overwrite paths and substituting a derived outfit for garbage-naked snapshots, backed by a quiet, event-driven, network-silent, CPU-deduped owner self-heal. It deletes the old per-frame `RunNakedNPCBugChecks` backstop.

---

## Reviewer checklist / things to verify

1. **Server guard correctness.** `ApplyActorData` ignores a naked overwrite only when `Content.HasWornItems() && !IsDead`. Confirm a legitimate pickpocket-to-empty still works (it goes through the incremental path, not here) and that a dead NPC loot still clears.
2. **EquipHook gate.** The new `&& !ScopedEquipOverride::IsOverriden()` applies to LOCAL actors. Confirm player/AI equips are not accidentally suppressed (they are not wrapped in the override) and that Vampire Lord cosmetic armor / debug-view equips still sync as before.
3. **`DeriveOutfitInventory` template walk** (`5c40ef7c`). Only walks to the template base when the leaf has no outfits at all; verify this doesn't double-derive for NPCs that intentionally have empty `outfits[0]` but a valid `outfits[1]`.
4. **Self-heal is main-thread only.** `0c9d20ce` removed the deferred cross-thread lambda. Confirm no remaining off-thread `SetActorInventory`/`Equip` call in the self-heal path.
5. **No orphaned references.** `RunNakedNPCBugChecks`, `EquipOutfit`, `IsWearingBodyPiece`, `ShouldWearBodyPiece`, `SetNakedDeadline`, `nakedDeadline`, `nakedLogged`, `SetOutfit` are all removed from this branch's tree (verified: 0 matches). `BGSOutfit.h` is retained in `Actor.cpp` because `DeriveOutfitInventory` legitimately uses `BGSOutfit*`.
6. **Perf regression check.** Confirm the leader no longer broadcasts a self-heal equipment storm (the dedup + `ScopedEquipOverride` gate). The 122 ownership cancels remain but should no longer cause re-dress churn.

---

# PR Review Details: NakedFix (improved naked-NPC fix)

**Branch:** `fix/Improved-naked-npc-fix` (absol89/TiltedEvolutionScriptFixes)
**Scope of this document:** ONLY the commits/code on `fix/Improved-naked-npc-fix`. It does NOT cover `diag/naked-npc-remote-skip` or `1.9-validation`; those are separate test branches. The older cosi bare `RunNakedNPCBugChecks` backstop is removed here (commit `2b6e44aa`) and is absent from this branch's tree.
**Why it exists:** The previous fix (PR #860, commit `f9909a23`) periodically polled every NPC and re-dressed any that looked naked. That only treated the symptom, ran every frame/second on the whole actor set, and introduced new bugs (crashes, network spam). This branch instead fixes the root cause at the snapshot/authority boundary and keeps a quiet, event-driven safety net.

---

## Problem statement

In Skyrim Together Reborn, NPCs that a player "owns" (nearest-client ownership) can render **naked on the owner's own screen** while looking dressed on every other player's screen.

Root cause: when ownership transfers or the owner reloads a cell, the NPC's authoritative inventory (`InventoryComponent::Content`) on the server is **wholesale-replaced** from the new owner's `GetActorInventory()` snapshot (`CharacterService::ApplyActorData`). During the transient undressed window after a cell reload — before the engine re-dresses the NPC — the owner's snapshot has no body piece, so `Content` becomes body-piece-less and that naked inventory propagates to every client. A second, independent source is the engine itself: when an NPC's 3D finishes loading, the engine issues a fully-naked `SetInventory` that can strip an already-dressed NPC.

Why the owner's screen specifically: remote clients dress NPCs from the spawn/assignment *snapshot*, but the owning client renders the NPC from its own (transiently naked) local state and relied on the ineffective periodic re-dress.

---

## The fix (commit by commit, in merge order)

### 1. `28637c65` — Server: ignore destructive fully-naked NPC inventory overwrites  **(root cause, server side)**
- In `CharacterService::ApplyActorData`, the authoritative `Content` was unconditionally overwritten by the incoming snapshot. Now it detects a **destructive fully-naked overwrite**: incoming snapshot has no worn entries (`IsFullyNaked()`) while `Content` still has worn items, and the NPC is alive. In that case the overwrite is **ignored** rather than applied.
- Legitimate fully-empty states are preserved: a dead NPC's naked overwrite is allowed (loot all), and pickpocket removals flow through the incremental `AddOrRemoveEntry`/`UpdateEquipment` path (which empties `Content` first), so they still sync. Only the racing naked overwrite on a living NPC is suppressed.
- Adds `Inventory::IsFullyNaked()` / `Inventory::HasWornItems()` (game-slot-agnostic helpers) in `Code/encoding/Structs/Inventory.{h,cpp}`.

### 2. `2b6e44aa` — Cleanup: remove unnecessary client naked-NPC backstop from PR #860
- Deletes `RunNakedNPCBugChecks()` and its `OnUpdate` call site in `InventoryService`, plus the old `IsWearingBodyPiece`/`ShouldWearBodyPiece`/`EquipOutfit`/`SetOutfit` helpers and `SetNakedDeadline`/`nakedDeadline` state. This PR replaces that approach entirely.

### 3. `0b558f28` — Client: substitute derived outfit for garbage-naked NPC snapshots
- `Actor::SetActorInventory` (client): if an incoming snapshot for a living, non-player NPC has no worn items, **derive the NPC's default outfit** via `DeriveOutfitInventory()` and substitute it instead of applying the naked snapshot. Non-worn contents sent by the server are kept (supplemented, not replaced). Properties/quest items handled via `SetInventoryRetainingQuestItems`.

### 4. `b32ee0e2` / `003bc0a9` / `d0ff2dbb` — Derive default outfit when snapshot is fully naked
- `Actor::DeriveOutfitInventory()` builds a dressed `Inventory` from `outfits[0]`/`outfits[1]`, resolving `LeveledItem` list A entries to their ARMO forms. `Actor.h` declares it. `PapyrusFunctions.cpp` drops a now-unused include.

### 5. `5c40ef7c` — Derive outfit from template base for FF-templated NPCs
- `DeriveOutfitInventory()` walks to the **template base** when the leaf base has no `outfits[0]`/`outfits[1]`. Without this, templated/leveled spawns (FF-prefixed) had no derivable outfit and stayed naked on the owner's screen. (Cosmetic: several `[NakedFix]` logs demoted from `info` to `debug` in `d0ff2dbb`.)

### 6. `04f06393` — Root-cause fix: skip fully-naked SetInventory strip of living NPCs (engine 3D-load undress)  **(root cause, client side)**
- In `TESObjectREFR::SetInventory`, when a living, non-player NPC receives a fully-naked inventory, the destructive `RemoveAllItems` strip is **skipped**. This is the engine's own 3D-load undress that was stripping already-dressed NPCs. Dead NPCs and genuinely changed inventories still apply normally. Wrapped in `ScopedInventoryOverride` so internal re-dress is consistent.

### 7. `ca0be5b0` / `9ec0526b` / `4c8fb0cd` / `0257de0b` — Owner self-heal safety net (client)
- `CharacterService::OwnerSelfHealDress()`: when an NPC becomes locally owned (and on 3D-apply / ownership reclaim), if its biped reports no worn armor, re-dress it from its own `GetActorInventory()` via `SetActorInventory` — the same proven REMOTE path, on the main thread, no deferred queue.
- Throttled (~1s) re-check for the late engine naked-overwrite; dresses immediately and re-checks next frame.

### 8. `024be397` / `bc369dfd` / `0c9d20ce` — Force re-equip worn armor + crash fixes + SetActorInventory redesign
- An earlier `ForceEquipWornArmor` approach caused a cross-thread crash and a broadcast storm (it ran off the network thread and called `EquipManager::Equip` directly). Redesigned (`0c9d20ce`) to reuse the existing REMOTE `SetActorInventory` path instead — no bespoke equip, no deferred cross-thread lambda. `bc369dfd` added the missing `ScopedEquipOverride` + load guard that the earlier version lacked.

### 9. `689ec519` — Kill self-heal CPU/network storms (dedup + suppress equip sync)  **(perf, critical)**
Measured on a 5-min playbench (build `0c9d20ce`): the leader ran **106 self-heal reapplies (69 unique actors)**, re-dressing the same NPC up to 10× due to MP ownership thrash (`CancelServerAssignment → respawn → reclaim`). Each reapply did `RemoveAllItems + full re-add + per-item Equip`, and for a LOCAL actor each per-item `Equip` fired `OnEquipmentChangeEvent` — the leader broadcast ~1000 equipment-change packets in 5 min, almost entirely self-inflicted. That was the CPU stutter.

- **Fix 1 (broadcast storm):** `EquipHook` already suppressed the sync for REMOTE actors under `ScopedEquipOverride`; extended the same gate to LOCAL actors (`EquipManager.cpp`) so internal re-dress equips don't broadcast. Wrapped the self-heal's `SetActorInventory` in `ScopedEquipOverride`. Engine equip still runs (NPC dressed locally); only the network sync is skipped (server already has the inventory). Player/AI-driven equips are NOT under the override, so genuine equip stays synced.
- **Fix 2 (CPU dedup):** skip the heavy re-apply when `GetEquipment()` already reports worn armor — already-dressed NPCs from an ownership reclaim no longer run `RemoveAllItems + re-add + re-equip`. Genuinely-naked NPCs (biped empty) still go through and get fixed.

---

## Net effect (verified in playbench logs)

Leader-client metrics before → after `689ec519` (same 10-min session, form-ID-agnostic assessment per project rules):
- Self-heal reapplies: **106 → 0**
- Self-initiated `sending equipment change` packets: **~1047 → 0**
- Ownership cancels (`CancelServerAssignment`) unchanged at 122 — that is vanilla MP assignment thrash and is **not** something this branch removes; the *consequence* (nakedness) is now guarded at both server and client.
- `SkipNakedStrip` guard (the `04f06393` 3D-load undress skip) fired 59× protecting 13 distinct NPC bases from being stripped.
- No code-level crashes attributable to this branch. (A separate SKSE crash observed was an MO2/usvfs environment fault, not this code.)

The naked-NPC symptom — local-owned NPC naked on owner screen, dressed elsewhere — is addressed by: server ignores the destructive naked overwrite (`28637c65`), client skips the engine 3D-load strip (`04f06393`), client substitutes a derived outfit for garbage-naked snapshots (`0b558f28`), and the quiet owner self-heal (`ca0be5b0`+) fixes any remaining local undress without spamming the network or CPU.

---

## Files touched (vs `origin/dev`)

```
Code/client/Games/PapyrusFunctions.cpp            |   2 -
Code/client/Games/Skyrim/Actor.cpp                | 228 +++---
Code/client/Games/Skyrim/Actor.h                  |   8 +-
Code/client/Games/Skyrim/EquipManager.cpp         |  23 +-
Code/client/Games/Skyrim/TESObjectREFR.cpp        |  16 +
Code/client/Services/Generic/CharacterService.cpp |  96 +++-
Code/client/Services/Generic/InventoryService.cpp |  85 +---
Code/client/Services/InventoryService.h           |   1 -
Code/encoding/Structs/Inventory.cpp               |  24 +-
Code/encoding/Structs/Inventory.h                 |   6 +
Code/server/Services/CharacterService.cpp         |  14 +-
11 files changed, 382 insertions(+), 121 deletions(-)
```

absol89:
[client1leadertruedressed.txt](https://github.com/user-attachments/files/29938707/client1leadertruedressed.txt)

Adding a log here from the latest commit test 41e3954, as previously couriers that only had a backpack slot could evade the checks. Checking for a body slot armor garment now seems to keep all npc from riverwood to whiterun dressed.

absol89:
Described the issue here, where derived outfits were writing all of the variants instead of just the picked final outfit https://github.com/absol89/TiltedEvolutionScriptFixes/issues/5

absol89:
https://github.com/absol89/TiltedEvolutionScriptFixes/actions/runs/32584707568 should have a failed playable win build

absol89:
Latest commit seems to fix the naked NPC, and the bandits outside helgen cave/standing stones didn't have duplicates, and the NPC in whiterun were not naked, and only 1 seemed to have some remnant of this bug (2 fine boots, not worn).

https://youtu.be/dg4ca4xRa5I

Playtestable build here https://github.com/absol89/TiltedEvolutionScriptFixes/actions/runs/32914592576

Log of the latest test 
[tp_client.log](https://github.com/user-attachments/files/31444748/tp_client.log)

miredirex:
gpt-5.6-sol max research on the problem:

> <...>
> I’ve identified the two attempts as PRs 860 and 880. Their own test history is useful because it confirms repeated
  ownership flips and transient “worn flags versus biped” disagreement, but both branches keep adding guards/self-healing
  around the symptom; I’m inspecting their exact patches before separating evidence from their proposed fixes.
> 
> PR 860 correctly identifies initialization/ownership traffic as dangerous, adds transition gating, and replaces the
  destructive global reset with a per-actor delayed fallback. However, `TakeOwnership()` has no `WaitingForAssignment` phase,
  queued events can cross the guard boundary, and the server still accepts claims and inventory/equipment traffic from
  arbitrary clients. It reduces exposure but does not establish single authority.
> 
> PR 880 removes `RunNakedNPCBugChecks()`, rejects or substitutes naked snapshots, derives default outfits, adds owner
  self-healing and retries, and eventually rebuilds actor 3D. Those are increasingly elaborate symptom guards. Its own
  measurements are particularly revealing: 106 self-heals across 69 actors, one NPC redressed up to ten times, and roughly
  1,000 equipment packets in five minutes due to ownership thrashing. Default-outfit derivation is also not authoritative
  for scripted, stripped, modded, or leveled outfits, while “ignore naked overwrite” does nothing against stale but non-
  empty equipment sets.
> 
> **Neither PR adds the missing owner/epoch validation or changes the direction of state transfer during handoff.**

\
research conclusion (pasting tldr here):

> ## Conclusion
> 
> The ownership-contention theory is strongly supported by the code. The most plausible root class is split-brain actor
  authority during loading and cell handoff:
> 
> - More than one client can temporarily behave as the local owner.
> - Ownership messages carry no generation or sequence.
> - The server accepts inventory/equipment changes from non-owners.
> - A newly loading claimant’s potentially incomplete inventory becomes authoritative.
> - Stale equipment events can clear the server’s worn flags or force another destructive re-equip pass.
> ## Recommended robust direction
> 1. Enforce authority immediately.
> 2. Make ownership server-granted and versioned.
> 3. Reverse the handoff state direction.
> 4. Reduce ownership churn.
> 5. Apply equipment once, after readiness.
> 6. Instrument before deleting the workaround.
> 
> The deletion criterion should be structural: exactly one accepted owner epoch, zero accepted old-owner inventory/equipment mutations, and no full inventory replacement caused merely by ownership transfer. At that point `RunNakedNPCBugChecks()` and its body-piece helpers can be removed confidently.
> 
> The authority defects date back to v1.0. The December 2023 “rare inventory mismatch” change (308353bc) added claimant snapshots to ownership transfers, likely amplifying the issue; the timer workaround arrived in March 2024. No files were modified during this investigation.

\
Given that your PR is essentially an amplified v2 of `RunNakedNPCBugChecks()` that doesn't address the underlying ownership problem, this isn't something I can merge as is. A proper fix shouldn't need to check "is this NPC wearing anything?" at all.

miredirex:
Closing in favor of #887

PR #881, opened 2026-07-18
Fix/issues#810 Where enemies going to local from remote got stuck in sheathe unequip state

### LINKED ISSUE #810: "Unequip" and "combatStanceStop" animation spam loop affects NPC behavior
When the party leader exits a cell, the other player receives ownership of all NPCs in that cell, and with that calls [`SetRemote(false)`](https://github.com/tiltedphoques/TiltedEvolution/blob/dev/Code/client/Services/Generic/CharacterService.cpp#L143) on every one of them [which unlocks NPC AI](https://github.com/tiltedphoques/TiltedEvolution/blob/187d9ede096aecce8ea25a3435e629a209eac94d/Code/client/Games/Animation.cpp#L28) behavior.

Immediately afterward, most if not all NPCs begin spamming "Unequip" and "combatStanceStop" (and possibly others) on themselves. This animation spam sometimes prevents NPCs from attacking players, since they are stuck repeatedly attempting to "Unequip" with no effect. Once the party member(s) follow the leader into the new cell, everything usually returns to normal (as AI is disabled again).

For debugging, you can insert `spdlog::info("Running action \"{}\" on {:X}", action.EventName.c_str(), action.ActorId)` [on this line](https://github.com/tiltedphoques/TiltedEvolution/blob/187d9ede096aecce8ea25a3435e629a209eac94d/Code/client/Games/Animation.cpp#L61-L62) to see the animation spam for yourself.

More context and potential solution:
<details>
<summary>Screenshot from #coding</summary>

</details>

absol89:
Fixes tiltedphoques/TiltedEvolution#810 after testing the following: [Video of fix in action](https://youtu.be/bYt5zKzvHsw)

# Code Review Summary — `fix/issues-810`

## Problem

When a party leader leaves a cell, remote NPCs get their AI localized on another
client (`SetRemote(false)`). Two bugs followed:

1. The engine replays a burst of reset/stance actions that got spammed over the
   network, freezing the NPC so it refused to aggro.
2. Even once unfrozen, a transferred hostile NPC would follow/detect a player but
   never draw or attack — it had no valid combat target and its combat controller
   was never started.

## Net change

5 files, +151 / −15. Two independent fixes.

### 1. Anti-freeze: dedupe the post-localize action burst

- **`ActorExtension.h`** — adds `LocalizedTick` (grace-window timestamp),
  `BroadcastedUnequip` / `BroadcastedCombatStanceStop` flags, and
  `kAnimReconcileGraceTicks = 60` (~1s).
- **`Animation.cpp`** — `HookPerformAction` now performs every reset action
  **locally** (visuals stay correct) but **broadcasts each reset type only once**
  within the grace window, instead of the old spam. `IsResetAction()` filters the
  relevant events (`Unequip`, `combatStance*`, `Sheathe/Draw`, `Idle`).
- **`CharacterService.cpp`** — stamps `LocalizedTick` and resets the broadcast
  flags at each localize site.

### 2. Combat engagement after transfer

- **`CombatService.cpp` / `Actor.cpp`** — on the engine's own detection hook
  (`HookUpdateDetectionState` — the vanilla "NPC noticed you" moment), if a
  **local, non-player NPC** detects a **player (local OR remote)** and it
  **arrived hostile**, it calls `StartCombatEx(player)` to boot the combat
  controller.
- **`ActorExtension.h` / `CharacterService.cpp`** — `ArrivedHostile` is captured
  at localize from `actorState.IsWeaponDrawn()` (transferred hostiles arrive
  weapon-drawn; peaceful NPCs arrive sheathed), used as the safety gate so
  friendly-area NPCs are never force-aggro'd.

## Design decisions worth a reviewer's eye

- **"Any player" not "remote player":** the target check accepts local *or* remote
  players. After transfer the NPC localizes on the *other* machine, where the
  discovered player is the **local** player — a remote-only check would miss the
  common case.
- **Detection, not `IsInCombat()`, is the trigger:** a transferred NPC's combat
  flag is never engine-initialized, so gating on it was dead. The engine's
  detection hook is the vanilla trigger and is used instead.
- **`ArrivedHostile` may be slightly strict** — in one playtest a sheathed bandit
  still engaged, hinting the gate could be looser. Left as-is since it works;
  first thing to loosen if an idle hostile ever refuses to aggro.

## Validation

Playtested: hidden → no premature discovery; approach after localize → bandit
draws and hits correctly; no interference with other enemies.  [Video of fix in action](https://youtu.be/bYt5zKzvHsw)

Logs:
[clientremote.txt](https://github.com/user-attachments/files/30153982/clientremote.txt)
[clientleader.txt](https://github.com/user-attachments/files/30153983/clientleader.txt)

### DIFF
--- a/Code/client/Games/ActorExtension.h
+++ b/Code/client/Games/ActorExtension.h
@@ -27,6 +27,13 @@ struct ActorExtension
     // and get stuck, refusing to aggro). 0 means "no grace window active".
     uint64_t LocalizedTick = 0;
 
+    // After a (re)localize, records whether the NPC arrived already in a combat-ready
+    // state (IsWeaponDrawn == true from the previous owner). Non-hostile NPCs arrive
+    // sheathed (false); transferred hostiles (bandits, etc.) arrive drawn (true). Used by
+    // the detection hook to engage combat ONLY for NPCs that were already hostile, so
+    // peaceful NPCs in friendly areas are never force-aggro'd.
+    bool ArrivedHostile = false;
+
     // During the reconcile grace window we still perform every action locally (so visuals stay
     // correct) but only broadcast each distinct reset action ONCE. These two flags track which
     // of the two loop actions we have already forwarded, so the server sees a single clean
--- a/Code/client/Games/ActorExtension.h
+++ b/Code/client/Games/ActorExtension.h
@@ -27,13 +27,6 @@ struct ActorExtension
     // and get stuck, refusing to aggro). 0 means "no grace window active".
     uint64_t LocalizedTick = 0;
 
-    // After a (re)localize, records whether the NPC arrived already in a combat-ready
-    // state (IsWeaponDrawn == true from the previous owner). Non-hostile NPCs arrive
-    // sheathed (false); transferred hostiles (bandits, etc.) arrive drawn (true). Used by
-    // the detection hook to engage combat ONLY for NPCs that were already hostile, so
-    // peaceful NPCs in friendly areas are never force-aggro'd.
-    bool ArrivedHostile = false;
-
     // During the reconcile grace window we still perform every action locally (so visuals stay
     // correct) but only broadcast each distinct reset action ONCE. These two flags track which
     // of the two loop actions we have already forwarded, so the server sees a single clean
--- a/Code/client/Games/ActorExtension.h
+++ b/Code/client/Games/ActorExtension.h
@@ -34,6 +34,19 @@ struct ActorExtension
     // peaceful NPCs in friendly areas are never force-aggro'd.
     bool ArrivedHostile = false;
 
+    // Issue #810: one-shot latch for the detection-hook combat engage. Once we have fired
+    // StartCombatEx for this NPC->player pair we set this true and NEVER fire again for the
+    // life of the actor. Without it the hook re-kicks StartCombatEx on every detection
+    // re-evaluation; StartCombatEx internally does StopCombat()+StartCombat(), and the
+    // StopCombat() sheathes the weapon -- producing the visible draw/sheathe OSCILLATION
+    // seen when a hostile NPC's ownership bounces between two nearby players (Embershard
+    // bandits looping). GetCombatTarget()!=target is NOT a sufficient latch because
+    // StopCombat() clears the combat target, so the next eval re-qualifies. This flag is
+    // the real latch: engage once, then leave all further combat/sheathe decisions to the
+    // engine. It is intentionally NOT reset on re-localize -- a contested NPC must not
+    // re-trigger the burst each time it flips back to local.
+    bool EngagedFromDetection = false;
+
     // During the reconcile grace window we still perform every action locally (so visuals stay
     // correct) but only broadcast each distinct reset action ONCE. These two flags track which
     // of the two loop actions we have already forwarded, so the server sees a single clean
--- a/Code/client/Games/ActorExtension.h
+++ b/Code/client/Games/ActorExtension.h
@@ -22,13 +22,20 @@ struct ActorExtension
     size_t GraphDescriptorHash = 0;
 
     // Tick (World clock) at which this actor was last (re)localized via SetRemote(false).
-    // Used by Animation.cpp to suppress the brief burst of reset/un-equip actions the engine
+    // Used by Animation.cpp to collapse the brief burst of reset/un-equip actions the engine
     // replays right after AI is unlocked (see issue #810: NPCs spam "Unequip"/"combatStanceStop"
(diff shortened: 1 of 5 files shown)


absol89:
Note that it's a bit hilarious in its eccd890 state, if you walk into certain areas you need to sheathe your weapons to make NPC abandon hostility. This can be seen when you enter the basement of Jorrvaskr the first time, or after fighting the vampires outside fort Dawnguard. Presumably this is because they were having weapons out, and fighting npc did StartCombat(player). Checking this to make a more thorough consequence analysis for the main quest. This quirk can be reprod in a solo game.

I will harden the checks to only apply to Aggression >=2 npc, and not trigger more than once onHit so it won't get intrusive

Test results on 64122b2 concluded that 3 wolves and 2 out of 3 bandits attacked party member after leader left the area. Jorrvaskr no longer triggered the startcombat(player) because it now requires the enemies to have an Agression AV of >=2

Code/client/Games/Skyrim/Actor.cpp
```diff
@@ -1172,6 +1172,50 @@ void TP_MAKE_THISCALL(HookUpdateDetectionState, ActorKnowledge, void* apState)
                 spdlog::debug("Cancelling detection from remote player to local player, owner: {:X}, target: {:X}", pOwner->formID, pTarget->formID);
                 return;
             }
+
+            // Issue #810 / #741: a local NPC whose ownership was transferred has no valid
```
miredirex, line 1176: Compact this comment

Code/client/Games/ActorExtension.h
miredirex, line 1: `ActorExtension` should not be extended with new fields unless absolutely necessary - and this isn't the case. If you want to store data on NPCs, use entt

Also please compact the comments to 1-2 lines after you move the fields

Code/client/Games/Animation.cpp
```diff
@@ -20,6 +20,20 @@ static TPerformAction* RealPerformAction;
...
+static bool IsResetAction(const TiltedPhoques::String& aEventName) noexcept
+{
+    return aEventName == "Unequip"
+        || aEventName == "combatStanceStop"
+        || aEventName == "combatStanceGo"
+        || aEventName == "Sheathe"
+        || aEventName == "Draw"
+        || aEventName == "DrawEnd"
+        || aEventName == "Idle";
+}
```
miredirex, line 35: Checking for exact event names and filtering them doesn't address the root cause of the bug

Code/client/Services/Generic/CombatService.cpp
```diff
@@ -162,41 +162,47 @@ void CombatService::OnNotifyProjectileLaunch(const NotifyProjectileLaunch& acMes
 
 void CombatService::OnHitEvent(const HitEvent& acEvent) const noexcept
 {
-#if 0
     if (!m_transport.IsConnected())
```
miredirex, line 165: This function was `#if 0`'d for a reason (#633), you essentially uncommented very old (and likely broken) combat code

Code/client/Games/Skyrim/Actor.cpp
```diff
@@ -1172,6 +1172,50 @@ void TP_MAKE_THISCALL(HookUpdateDetectionState, ActorKnowledge, void* apState)
...
+            const auto* pTargetEx = pTargetActor->GetExtension();
+            if (pOwnerEx->IsLocal() && !pOwnerEx->IsPlayer() &&
+                (pTargetEx->IsLocalPlayer() || pTargetEx->IsRemotePlayer()) &&
+                pOwnerEx->ArrivedHostile && !pOwnerEx->EngagedFromDetection &&
+                pOwnerActor->GetActorValue(ActorValueInfo::kAggression) >= 2.0f)
+            {
+                if (pOwnerActor->GetCombatTarget() != pTargetActor)
+                {
+                    spdlog::info("Combat started (detection): local NPC {:X} -> player {:X} (remote={})", pOwner->formID, pTarget->formID, pTargetEx->IsRemotePlayer());
+                    pOwnerActor->StartCombatEx(pTargetActor);
```
miredirex, line 1215: See `Actor::StartCombatEx`'s fn comment

miredirex (changes requested):
Vibecoding bug fixes is not the way. If you want to delegate a bug fix to an LLM, you need to understand why the bug occurs in the first place

absol89:
I have experimented with other tweaks (more of a feature), and couldn't conclude that the commits in fix/issues-810 resolved the core issue. At least not without workarounds that deviate from vanilla agression behaviors. Closing the vibed bugfix pr.

PR #875, opened 2026-06-17
Feature: Respawn overrides system for interior cells deemed problematic to die in

absol89:
Fixes tiltedphoques/TiltedEvolution#878
In regards to the bug, I fixed it with a 1.5 second fallback knockdown during the blackscreen. Works indoors+outdoors.
https://youtu.be/SoOoVqPw-bw

This feature adds a RespawnOverrides.cpp file where players can modify respawn locations through CellEditorId names.
This was later refactored to create an RespawnOverrides.ini file in the config folder, supports patching positions locally.

It runs once every player death, so performance is not more important than convenience of editing or avoiding formIDs.

It of course supports all vanilla locations (special edition+dlc) but can also be extended with modded/cc interiors safely.

Tested these in https://github.com/absol89/TiltedEvolutionScriptFixes/actions/runs/28106445654 before final correction.
Example: Overriding the ini values of Korvanjund02 to Korvanjund02=1201, -636, -400 results in a better spawn.
image

Safeguards: If the player spawns under the floor somehow by physics or misconfigured positions, they teleport to COC.

Comment: While respawning code was originally in PlayerCharacter.cpp i refactored it to make it less merge conflictable.

Steps: Note that the original coc command will put you at the OLD 1.8.0 spawn location and player.kill at the new spawn.
coc QASmoke
connect to localhost server
player.kill
coc HaltedStreamCamp01
player.kill
coc Korvanjund01
player.kill
coc Korvanjund02
player.kill
coc Saarthal02
player.kill
coc ShimmermistCave02
player.kill
coc StillbornCave01
player.kill
coc SwindlersDen01
player.kill

Before: player used to respawn in a broken position or too close to trappable locations or enemies or boss rooms/chests.

After: player respawns closer to the initial entrance of the cell, instead of the center of cell (which is used for non-list cells)

Example: Dying in Saarthal02 would originally place you at the boss where the Eye of Magnus is, skipping dungeon areas.
image

Before fix: Player respawns near Eye of Magnus
image

After fix: Player respawns at Saarthal02 entrance
image

PR #873, opened 2026-06-16
Fix/issue 871 restore password hiding parity in connect dialogue

absol89:
Fixes tiltedphoques/TiltedEvolution#871 by copying logic from the server browser's connect-password.component to the connect.component which is more used. New behavior is that the password is censored by default, and it can be toggled.

image

verified 16 june in https://github.com/absol89/TiltedEvolutionScriptFixes/actions/runs/27628058512

PR #889, opened 2026-09-06
v1.8.1

miredirex:
### Full changelog
- Added support for the latest Skyrim SE and SKSE versions (1.7.104, 2.2.6+) #885
- Synced leveled NPCs between clients – no more desynced animal types or randomized bandit appearances in encounters #892
- Refined and stabilized the NPC ownership transfer system, fixing several issues including NPCs appearing naked #887
- Fixed dragons occasionally failing to spawn for party members 92980e22
- Enabled party leaders to use the '/settime' chat command on private servers #877
- Fixed dialogue sync when the conversation participant was a party member instead of the leader #896
- Fixed a rendering bug that previously caused Community Shaders' Upscaler to crash #829
- Fixed an issue where using the Sanguine Rose or Staff of the Familiar could summon two creatures that fought each other #894
- Added an ESL-flagged SkyrimTogetherQuestPatches.esp; currently contains one patch for the Companions questline #855
- Fixed a rare bug where camera would get stuck after a respawn 2e52a704
- Disabled party auto-join on public servers #874
- Added respawn coordinate overrides for several problematic interior cells / dungeons #875
- Fixed an animation bug with vampire lord transformation #831
- Fixed internal directory path detection that sometimes caused Community Shaders (and possibly other mods) to malfunction #828
- Added "Clear stuck screen effects" to the F3 helpers menu to remove persistent effects such as blindness #850
- Fixed players sometimes getting stuck in the wood-chopping animation (`ResourceFurnitureScript` patch) #841
- Added various UI quality-of-life tweaks and improvements #868, #873
- Fixed a rare startup crash related to CEF #844
- Added "incompatible game version" popup if mismatch is detected 24dc4c4e
- Removed sentry and crashpad to reduce antivirus false positives #851
- Several CI/CD improvements


Force67:
Read through the whole diff against `master` (180 files, head `5aec348`). The ownership epoch scheme is the right shape. A monotonic counter per `OwnerComponent`, validated server-side on every mutating request, beats the old "first client to ask wins" model. Moving the authority checks into `CanClaimOwnership` and `TransferOwnership` also puts the rules in one readable place. Notes below, heaviest first.

## Fix before merge

**`Code/server/Services/CharacterService.cpp:519` dereferences a possibly-null `try_get`.**

```cpp
m_world.try_get<AnimationComponent>(*it)->ActionsReplayCache.Clear();
```

The view above it is `<OwnerComponent, CharacterComponent>`, so `AnimationComponent` is not guaranteed, and `acMessage.Packet.ActorId` comes straight off the wire. `PopulateAssignmentResponse` in the same file uses the checked form. Either `get<>` it (if every character entity carries one) or null-check it.

**`Code/client/CrashHandler.cpp:124` calls `spdlog::shutdown()` on a path that can resume the process.**

This replaced the earlier `flush()`. A few lines up, the handler forwards to a third-party unhandled-exception filter and sets `retval = EXCEPTION_CONTINUE_EXECUTION` when that filter says to continue. The game then keeps running with the whole spdlog registry dropped and the async thread pool gone. Every later log line is lost, and a second crash records nothing. A flush achieves what the change wanted without tearing down logging.

**`Code/client/Services/Debug/DebugService.cpp:298` mutates the list it is iterating.**

```cpp
for (const auto& modifier : TES::Get()->activeImageSpaceModifiers)
    ImageSpaceModifierInstance::Stop(modifier.object);
```

`Stop` (18563) removes the instance from `TES::activeImageSpaceModifiers`, and `GameValueList::Iterator::operator++` then reads `next` from an entry the engine may have unlinked or freed. Because the head entry is embedded by value, removal of the first element rewrites the node the iterator is standing on. Copy the pointers into a local `Vector` first, then stop them. `modifier.object` also wants a null check.

## Worth a second look

**`Code/server/Services/CommandService.cpp:16` declares a third `bAnnounceServer`.**

`GameServer.cpp:35` and `ServerListService.cpp:19` already declare the same setting name. All three pick up the same ini value at load, but `/set LiveServices:bAnnounceServer true` at runtime only reaches whichever instance `FindSetting` returns first, so the `/settime` gate and the actual listing behaviour can disagree. This PR adds `GameServer::IsPublicServer()` for this exact question; call it instead.

**`ProcessLeveledConforms` has no way out of `WaitingForDisable` (`CharacterService.cpp:1663`).**

`WaitingFor3D` gives up when the cell detaches, but a `DisableImpl()` that never lands keeps the entry pending and the actor disabled forever. `OnDisconnected` re-enables, `OnActorRemoved` only resets the stage. A wall-clock or frame deadline that re-enables and drops the entry would bound the worst case (an invisible NPC for the rest of the session).

**`VersionDb::LoadV5` (`VersionDb.h:295`) trusts `addressCount` and never checks the stream.**

A truncated or corrupt versionlib runs the loop up to 2^32 times with two map inserts each. `LoadV2` bounds `tnLen`; a `file.good()` check in the loop would match that.

**`ShiftGridCellRequest::Cells` is now dead payload.** `PlayerService.cpp:74` switched to `IsInRange`, so the server never reads the list, but the client still serializes every cell on each grid shift. The opcode table already breaks compatibility this release, so dropping the field is free.

**`OnOwnershipTransferRequest` passes `OwnershipTransferReason::Relinquish` for both release reasons**, so a declined grant logs as "owner relinquished control". Cosmetic, but it discards the distinction `OwnershipReleaseReason` was added to carry.

**`Code/tests/encoding.cpp` lost the `CancelAssignmentRequest` case and gained nothing.** Every message that grew an `OwnershipEpoch` still round-trips with 0 on both sides, so the new varint is never exercised. Setting a non-zero epoch on two or three of them would cover the family.

## Cleanup

**Dead hooks merged in.** `Code/client/Games/Skyrim/Components/TESActorBaseData.cpp` is an entire file whose `Initializer` body is commented out. `TESObjectREFR.cpp:1133` keeps `HookSetLeveledCreature` with its pointer, assignment, and `TP_HOOK` all commented out. Both say "Kept for reference"; git is that reference. `TESImageSpaceModifier.h` and `ImageSpaceModifierInstanceForm::Stop` have no callers either (DebugService uses the base overload, which the derived static hides).

**`RenderSystemD3D11` constructor is now four comment lines** (`RenderSystemD3D11.cpp:18`). `m_createConnection`, `m_renderConnection` and `m_resetConnection` survive in the header with nothing writing them, so they are uninitialized `size_t` members. `OnReset` has no caller left.

**Three new open-question TODOs.** `ActorExtension.h:31` ("atomic? bool instead? maybe simplify to `IsReenabling()` ?"), `Primitives.h` ("Migrate from GameList to this (and rename)"), `DiscoveryService.cpp:249` ("GetById performance in loop?"). Issues hold open questions better than headers do.

**`ApplyLeveledNpcPick` logs at info on its no-op paths** ("already matches owner's pick", "whose base is not a leveled temp"). Every remote NPC in a loaded cell hits one of those on assignment. debug fits.

## Build and release

**`Code/client/xmake.lua:75` dropped the `v0.2.7` pin on tiltedcore.** A released client can no longer be rebuilt from its tag; whatever the repo serves at build time is what ships.

**The release path has never run.** `release.yml` and `docker-server-image.yml` only trigger on tags, schedule, or dispatch, so the first tag push is also the first test. The asset names have to match exactly across three files: `windows.yml` derives the version from `git describe --tags --always`, while `linux.yml` and the `Verify release assets` step derive it from `${GITHUB_REF_NAME#v}`. If the tagged commit carries a second tag, describe can return the other one and the `test -f` check fails after a full Windows build. Using `GITHUB_REF_NAME` in all three removes the class of failure.

**Dockerfile moved off distroless** to `debian:12-slim`, with "(todo: maybe reconsider 'distroless' in the future)". What forced it? Removing crashpad should have made the distroless image simpler, not harder. Separately, `docker-compose.yml` remapped the volumes from `/home/server/...` to `/st-server/...`; existing deployments silently stop persisting config and logs unless the release notes call that out.

**`ResourceFurnitureScript.pex` is the first vanilla script this repo overrides as a loose file.** Any other mod shipping that script wins or loses on load order, and the compiled .pex cannot be checked against the .psc in review. Worth a wiki line on how to rebuild it.

PR #882, opened 2026-07-18
Sync leveled NPCs identity across clients

absol89:
### Overview

This PR adds synchronization for leveled NPC identities so the same actor resolves consistently across Skyrim Together clients.

Because Skyrim evaluates leveled-character lists locally using player level, list rules, and RNG, clients can resolve the same Ref ID to different NPCs. This can cause mismatched race, sex, model, face, skeleton, or animation behavior.

The selected NPC Base ID is now treated as shared actor state. The first valid identity becomes canonical and is preserved across assignment, spawning, observation, and ownership transfer.

---
---
Fixes #639 and #338

<details>
  <summary>Details (mostly outdated)</summary>
# PR: `feat/leveled-npc-sync-v2` vs `upstream/dev`

**Branch:** `absol89/TiltedEvolutionScriptFixes@feat/leveled-npc-sync-v2` → `tiltedphoques/TiltedEvolution@dev`
**Merge base:** `9d81ef07` (current `upstream/dev` HEAD, so the branch is rebased on upstream dev)
**Scope:** 16 files changed, +325 / −19

## Goal

Make every client render the **same leveled-NPC identity** that the actor's *owner* rolled, instead of each client independently rolling its own pick from the leveled list. Without this, two players standing near the same leveled bandit see different NPCs (different faces/races/equipment), and ownership transfers can leave remote clients stuck on the wrong pick.

## How it works

The owner captures the resolved pick at assignment time and the server stores/relays it. Every other client conforms its locally-rolled actor to the owner's pick by swapping the base form (disable → re-enable 3D → swap base).

### Wire protocol — new `LeveledNpcPickId` field
Added to four messages (client ↔ server):
- `AssignCharacterRequest` (.h/.cpp)
- `AssignCharacterResponse` (.h/.cpp)
- `CharacterSpawnRequest` (.h/.cpp)
- `NotifySpawnData` (.h/.cpp)

Each gets the new `GameId LeveledNpcPickId{}` field plus serialize/deserialize and `operator==` updates.

### Server side (`Code/server/`)
- `CharacterComponent.h`: new `FormIdComponent LeveledNpcPickId{}` stored on the character.
- `CharacterService.cpp`:
  - `CreateCharacter`: stores `message.LeveledNpcPickId` as received (client-authoritative, same trust model as `BaseId`).
  - `Serialize` (spawn): copies the stored pick into `CharacterSpawnRequest::LeveledNpcPickId`.
[...]

PR #892, opened 2026-09-12
Sync leveled NPCs identity between clients

### LINKED ISSUE #338: Implement recycleactor to recreate persistent randomised NPCs to ensure synchronisation of base form IDs
Hand placed NPCs synchronise across clients, however the random outcome of the selection from levelled lists is not taken into account. Gender, race and species may appear different across clients.

We might need to reverse engineer the recycleactor function in order to convert the NPCs to the correct base ID on other clients who spawn them in as suggested by @RobbeBryssinck

### LINKED ISSUE #639: Leveled Actor baseref sync
## Description

Even with the lastest version (1.6.5), leveled actors (bandits, foxes, bears, etc) can spawn with different baseref on each clients due the fact that its a list of possible monster/animals based on the player level and RNG.

### One potential fix is the following:

OnCellLoad ->

**Case: Party Leader loads the cell after someone else loaded the cell**

Compare loaded leveled actors baseref with the one the party leader wants to spawn. If they match, don't do anything. If they don't, disable existing ones and spawn party leader ones.

**Case: Non Party Leader loads the cell after someone else loaded the cell**

Compare loaded leveled actors baseref with the one the party leader already spawned. If they match, don't do anything. If they don't, disable existing ones and spawn the party leader ones.

## Additional Notes 

﻿**what was it again with Leveled Actors. Level based spawning was synced right?**

yes 
except when it doesnt
if a troll is disabled for level 1 players, but enabled for level 10 players, the troll will appear for both players
if an actor can be a troll or a bear depending on the player's level, there will probably still be desync

but if you're roughly the same level, it shouldn't matter too much
plus, something will appear regardless
so far, i haven't found it to be a big deal
usually, it's just a skeever on one client, and a wolf on another
both take like 3 hits, so its kinda whatever
the health and damage of the entity will still be synced

**so youre fighting a skeever with the health of a wolf?**

basically, yeah
depending on who owns the actor

The host of that actor determines its inventory
Regardless of leveled actor mismatches
This is why you might find skeever tail on what looks like a bear for you

https://www.reddit.com/r/SkyrimTogether/comments/ee9n73/i_see_a_fox_friend_sees_a_bunny/

Potentional quest where this could be a **visual** issue https://elderscrolls.fandom.com/wiki/Animal_Extermination

miredirex:
### Overview
This PR adds synchronization for leveled NPC identities so the same actor resolves consistently across Skyrim Together clients.

Because Skyrim evaluates leveled-character lists locally using player level, list rules, and RNG, clients can resolve the same Ref ID to different NPCs. This can cause mismatched race, sex, model, face, skeleton, or animation behavior.

The selected NPC Base ID is now treated as shared actor state. The first valid identity becomes canonical and is preserved across assignment, spawning, observation, and ownership transfer.

Closes #639
Closes #338

(v1 implementation: 4fb53dd...200c5d4, v2 #882, this is v3)


SlayerTheChikken:
This is huge, thank you!

PR #893, opened 2026-09-13
Exclude faction jail containers from object sync (#700)

### LINKED ISSUE #700: Inventory glitch when thrown in jail
Whenever the host goes to jail and it brings the other player in the party with them and they serve their time the other players inventory gets overridden by the host players inventory with the exception of equipped items.

sfedev:
Fixes #700

## What was wrong

When a player is sent to jail, the game moves their whole inventory into the jail's prisoner belongings container (stolen goods go to the evidence chest) and hands it back when the sentence is served. Both are regular world containers, so `ObjectService::OnCellChange` registered them on the server with the first player's stash as their inventory. When a second party member entered the same jail cell, the server answered with that inventory and `OnAssignObjectsResponse` replaced the second player's local chest contents with the leader's items. On release, the game gave the second player the leader's belongings and their own were gone.

Example: leader and member both get arrested in Whiterun and serve their time. Before this change, the member walks out with the leader's inventory (except worn items). After it, each player gets their own items back.

Two quest chests (`0x39CF1` in Diplomatic Immunity and `0x3EF03` in No One Escapes Cidhna Mine) were already excluded by hard-coded form ID for the same reason. As noted in the issue, that does not scale: every hold has its own jail chests, and mods add more.

## What changes

Every faction that can arrest the player stores, in its crime data, the container that receives the player's inventory (`PLCN`) and the one that receives stolen goods (`STOL`). On cell change the client now walks the loaded factions (`TESDataHandler` faction array) and excludes those containers from object sync, so each client keeps its own copy and the game returns each player's own items. The check compares pointers only and never dereferences the crime data, so a layout mismatch cannot crash the client; it would just fall back to the previous behaviour for that faction.

The two hard-coded quest chests stay excluded because they are quest-owned containers, not faction jail containers.

- `Code/client/Games/Skyrim/Forms/TESFaction.h`: crime data layout.
- `Code/client/Games/TES.h`: faction form array in the data handler, with offset asserts.
- `Code/client/Services/Generic/ObjectService.cpp`: collect the containers and exclude them in `ShouldSyncObject`.

## How it was checked

- `SkyrimTogetherClient` builds and `TPTests` pass on this branch (based on `dev`).
- In-game verification with two arrested party members is still pending. Expected result: each player gets their own inventory back, and the client log shows `Excluding sync for <chest id>` for the prisoner belongings and evidence chests when entering the jail cell, while other containers in the cell keep syncing.

Limited to object sync exclusion; no changes to actor inventories, ownership transfer, protocol or quest logic.

### DIFF
--- a/Code/client/Games/Skyrim/Forms/TESFaction.h
+++ b/Code/client/Games/Skyrim/Forms/TESFaction.h
@@ -3,7 +3,43 @@
 #include <Forms/TESForm.h>
 #include <Components/TESFullName.h>
 
+struct TESObjectREFR;
+struct BGSListForm;
+struct BGSOutfit;
+
+struct TESReactionForm : BaseFormComponent
+{
+    void* reactionListHead; // BSSimpleList<GROUP_REACTION*>
+    void* reactionListNext;
+    uint8_t groupFormType;
+    uint8_t pad19[7];
+};
+
+static_assert(sizeof(TESReactionForm) == 0x20);
+
 struct TESFaction : TESForm
 {
+    // Where the game sends the player's belongings when they are arrested by this faction.
+    struct CrimeData
+    {
+        TESObjectREFR* jailMarker;               // JAIL
+        TESObjectREFR* waitMarker;               // WAIT
+        TESObjectREFR* stolenGoodsContainer;     // STOL: stolen items taken on arrest
+        TESObjectREFR* playerInventoryContainer; // PLCN: the rest of the player's inventory, returned on release
+        BGSListForm* crimeGroup;                 // CRGR
+        BGSOutfit* jailOutfit;                   // JOUT
+        uint8_t crimeValues[0x18];               // CRVA
+    };
+
     TESFullName fullname;
+    TESReactionForm reactionForm;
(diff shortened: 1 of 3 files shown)


Code/client/Games/Skyrim/Forms/TESFaction.h
```diff
@@ -3,7 +3,43 @@
...
+        TESObjectREFR* stolenGoodsContainer;     // STOL: stolen items taken on arrest
+        TESObjectREFR* playerInventoryContainer; // PLCN: the rest of the player's inventory, returned on release
+        BGSListForm* crimeGroup;                 // CRGR
+        BGSOutfit* jailOutfit;                   // JOUT
+        uint8_t crimeValues[0x18];               // CRVA
+    };
+
     TESFullName fullname;
+    TESReactionForm reactionForm;
+    uint8_t crimeGoldMap[0x30]; // BSTHashMap<const TESNPC*, uint32_t>
```
miredirex: Where did you get these types and offsets from? `crimeGoldMap` is a [pointer to the map](https://github.com/powerof3/CommonLibSSE/blob/dev/include/RE/T/TESFaction.h#L193)

PR #897, opened 2026-09-15
Advance the shared time when a player sleeps or waits (#467)

### LINKED ISSUE #467: Use sleep / wait in Skyrim to actually progress time on the server
It is really non-immersive having to play through the night without the ability to skip time when all players go to sleep. You can roleplay an insomnia, but still...you will have to for example wait half an hour real-time for shops to open or disconnect from the server to sell items.

I saw a lot of requests for synced sleep, but I still haven't been able to find an open issue or any pull request here.
Is someone working on this feature? Is there a reason why not to?

I've found this "[Add TimeScale Server Setting](https://github.com/tiltedphoques/TiltedEvolution/pull/427)" merged. So I thought it could be imho easily used to progress time on server when all the clients notify the server that they are either waiting or sleeping.

sfedev:
Fixes #467

## What was wrong

While connected, the client locks the game clock (`HookSimulateTime`) and
`CalendarService` writes the server's hour, day, month and year every frame.
Sleeping or waiting therefore does nothing: the sleep/wait menu advances the
game clock and the next frame overwrites it. A party that wants to skip the
night has to disconnect, or the leader has to type `/settime` (#702, #877).

Example: two players in a party outside Riverwood at 22:00. Player A waits
8 hours. Before this change, both clients still show 22:00 afterwards and
the shops stay closed. After it, both clients jump to 06:00 and the day
advances for everyone.

## What changes

The client turns a finished sleep or wait into a request and the server
advances the shared clock for everyone, so the existing time sync stays the
single source of truth.

- `Code/client/Games/Skyrim/Events/EventDispatcher.h`: `TESSleepStartEvent`,
  `TESSleepStopEvent`, `TESWaitStartEvent` and `TESWaitStopEvent` get their
  fields (start time and desired end time in game days, `interrupted`), and
  the three dispatchers that were still unnamed are typed.
- `Code/client/Services/Generic/CalendarService.cpp` / `.h`: sinks for those
  four events. On start it records the requested hours; on stop, if the
  sleep/wait was not interrupted and the client is online, it sends
  `RequestTimeSkip`. When a server time update jumps forward, the difference
  is added to `GameDaysPassed`, as the game does when the player sleeps, so
  script timers keep working.
- `Code/encoding/Opcodes.h`, `Messages/RequestTimeSkip.{h,cpp}`,
  `Messages/ClientMessageFactory.h`: new client message with the hours to skip.
- `Code/server/Services/CommandService.{h,cpp}`: handles the request with the
  same permission rule as `/settime` (admins always; the party leader on
  private servers). `CanChangeTime` factors that rule out.
- `Code/server/Services/CalendarService.{h,cpp}` and
  `Code/common/DateTime.{h,cpp}`: `AdvanceTime(hours)` uses a new
  `DateTime::AddHours`, which is the day/month/year rollover that `Update`
  already did, and resends `ServerTimeSettings` to loaded players. Skips are
  limited to more than 0 and at most 48 hours.
- `Code/tests/encoding.cpp`: round trip of the new message.

Not changed: the time scale, `bSyncPlayerCalendar`, `/settime`, the clock
lock, fast travel, quests, AI or inventory. A member who is not the leader can
still open the sleep/wait menu; the server ignores the request and logs it at
debug level.

## How it was checked

- On this branch (`feat/467-sleep-wait-time-skip`, base `dev` `e0002073`):
  client, server, runner, early-load and TPProcess build with zero errors and
  `TPTests` pass (28 assertions in 5 test cases, the suite present in `dev`;
  the new round-trip test is one of them). On the fork's integration branch
  the same change also passes `TPServerTests` (145/9).
- In-game, two local clients on a fork build carrying this same change
  (15 September 2026): the leader waiting 3 h and sleeping 3, 2, 8 and 19 h
  moves both clients together, including day changes (server log:
  `advanced the shared time by 3 hours`, other client: `Shared time jumped
  forward by 0.125 days`); a non-leader's wait is ignored; after recreating
  the party the new leader can skip time; `/settime` and cell changes behave
  as before. The event times are confirmed to be game days.
- Fast travel while connected (four cell changes by two clients in a party,
  15 September 2026) sends no time skip and both clocks stay in sync: no
  sleep/wait event, no request and no server-side advance were logged.
- Not verified: scripted sleep interruptions are out of scope; the
  GameDaysPassed adjustment is a few game minutes off per skip (drift between
  the local mirror and the server clock), which no vanilla timer notices.

Limited to sleep and wait; does not touch fast travel or the time scale.


absol89:
Remember public servers can have several party leaders (against playguide) and players outside parties who may use this for trolling.

My previous implementation kept control to oldest party leader on the server, but wasnt able to guarantee that sleeping resting fatigue actually worked. Should be tested with survival

sfedev:
Thanks for the review.

The request goes through the same gate as `/settime` from #877
(`CommandService::CanChangeTime`): admins always, and a party leader only when
`bAnnounceServer` is false. On a public server a sleep or wait from anyone who
is not an admin is ignored (debug log only), so players outside a party or
extra party leaders cannot move the clock there. On a private server every
party leader can, which is the same as `/settime` today. If you prefer to
limit it to the leader of the oldest party, I would do that in a separate PR
so it applies to both `/settime` and this.

Survival Mode is not tested yet, you are right. The change does not interrupt
the sleep: the game completes it locally and only then the server advances
the shared clock by the same hours and the client adjusts `GameDaysPassed`,
so the game's own sleep-end handling runs as usual. It still needs an in-game
check with Survival enabled on both clients.

